PR #5, opened 2019-12-31
minor spelling tweaks

brettkoonce:


Sources/PublishCLICore/CLIError.swift
```diff
@@ -8,7 +8,7 @@ internal enum CLIError: Error {
     case newProjectFolderNotEmpty
     case notASwiftPackage
     case failedToResolveSwiftPackageName
-    case outputfolderNotFound
+    case outputFolderNotFound
```
dmonagle: Looks like you have renamed the enum to proper camel-case but not replaced various uses of it in the code. You can always use Xcode's Refactor->Rename to avoid issues like this in future. At this point the project won't build with this change.

brettkoonce: thanks for the tip!

PR #6, opened 2019-12-31
Modified the run command to allow for instant regeneration

dmonagle:
The run command will now regenerate every time enter is pressed until the user enters a single "q" (case insensitive) or presses control-c. This makes for much faster iterating when testing design changes.

As a side-note, part of the motivation for this was that I found about every second use of the "run" command led to the server seemingly starting correctly but I couldn't connect on port 8000. Rerunning usually resolved the error. While this continues to happen, the failure to connect never occurs when using the new feature to regenerate the site - only when restarting it.

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -12,17 +12,44 @@ internal struct WebsiteRunner {
     let folder: Folder
 
     func run() throws {
-        let generator = WebsiteGenerator(folder: folder)
-        try generator.generate()
-
-        let outputFolder = try resolveOutputFolder()
         let portNumber = 8000
+        
+        try self.generate()
 
         print("""
         🌍 Starting web server at http://localhost:\(portNumber)
-
-        Press any key to stop the server and exit
         """)
+
+        var quit = false
+        
+        while !quit {
+            try startServer(portNumber: portNumber)
+
+            print("""
+
+            Press ⮐  to regenerate the site.
+            Type q⮐  or control-c to quit
+            """)
+            let input = readLine(strippingNewline: true)!
+            switch input.uppercased() {
+            case "Q":
+                quit = true
+            default:
+                try self.generate()
+            }
+        }
+    }
+
+}
+
+private extension WebsiteRunner {
+    func generate() throws {
+        let generator = WebsiteGenerator(folder: folder)
+        try generator.generate()
+    }
+    
+    func startServer(portNumber: Int = 8000) throws {
+        let outputFolder = try resolveOutputFolder()
 
         DispatchQueue.global().async {
             _ = try? shellOut(
@@ -30,12 +57,8 @@ internal struct WebsiteRunner {
                 at: outputFolder.path
             )
         }
-
-        _ = readLine()
     }
-}
 
-private extension WebsiteRunner {
     func resolveOutputFolder() throws -> Folder {
         do { return try folder.subfolder(named: "Output") }
         catch { throw CLIError.outputfolderNotFound }


daniloc:
Finding the same issue with alternating connection failures. This sounds handy!

natebird:
I'm also running into the issue with alternating connection failures.

javb99:
What about auto restarting on file changes using something like [File System Events](https://github.com/eonil/FSEvents)?

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -12,30 +12,53 @@ internal struct WebsiteRunner {
...
         }
-
-        _ = readLine()
     }
-}
 
-private extension WebsiteRunner {
     func resolveOutputFolder() throws -> Folder {
         do { return try folder.subfolder(named: "Output") }
         catch { throw CLIError.outputfolderNotFound }
```
javb99:
```suggestion
        catch { throw CLIError.outputFolderNotFound }
```
This change did not compile for me.
This line has a typo I think. The 'F' in Folder is not capitalized.

javb99:
I also noticed the issue about restarting the server. I found the error by connecting the `shellOut` for the http server. This is the error I got:
```
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SimpleHTTPServer.py", line 235, in <module>
    test()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SimpleHTTPServer.py", line 231, in test
    BaseHTTPServer.test(HandlerClass, ServerClass)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/BaseHTTPServer.py", line 606, in test
    httpd = ServerClass(server_address, HandlerClass)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 420, in __init__
    self.server_bind()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/BaseHTTPServer.py", line 108, in server_bind
    SocketServer.TCPServer.server_bind(self)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 434, in server_bind
    self.socket.bind(self.server_address)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 48] Address already in use
```
Perhaps the server is not being closed properly?

ncreated:
> Perhaps the server is not being closed properly?

@javb99 I guess it is not, the `SimpleHTTPServer` process is still running after exiting from `publish run`. I kill it manually until it's fixed:

`pkill -f SimpleHTTPServer`

JohnSundell:
I'll review this PR as soon as I can, but just wanted to give everyone here an update that I'm aware of the issue that the Python server isn't properly shut down when exiting `publish run`, and that I have a fix in the works. I've personally always just left the server running until I restarted my machine, because I tend to work on my site almost every single day, which I guess is why I never encountered this myself 😅 Manually killing the process like @ncreated mentioned is a great workaround until a fix has been deployed 👍

dmonagle:
@JohnSundell While I made the fact that the python server doesn't shut down correctly a side-note, it really was only partly why I made this a pull request. I found that while tweaking styles, I was constantly switching to the terminal and pressing <up+enter> to regenerate. This made it a little faster, but in retrospect, I don't think this should be the solution.

As @javb99 mentioned, it would be great to have automatic reload on file change, but I also think that this is something outside of the scope of the command line tool. Both Linux and MacOS use different methods to do this and my previous experience is that this is always an ongoing battle to "get it right".

My preferred solution would be for us as a community to find one of the many file watching tools that is easy to install and post instructions for using it to rerun the `publish generate` or `publish run` commands of this tool. Then we can add it as an optional step in the README for people to reference. If I can find time to get this done, I'll submit it as a separate PR.

I'm curious that you mention that when developing the site you leave the python server running constantly. Are you just running the `publish generate` command from another terminal while leaving the python server running?

Lastly, back on the server issue, would you consider a PR that uses a [vapor](https://github.com/vapor/vapor) server to serve up the content? I know this would require an external dependency for the command line tool, but vapor has always been a super reliable web server and it has the advantage that you don't need to "ShellOut". Happy to put that together and raise a PR if you will consider it.

Thanks again for this extremely useful framework. ⭐️⭐️⭐️⭐️⭐️

ncreated:
> I'm curious that you mention that when developing the site you leave the python server running constantly. Are you just running the publish generate command from another terminal while leaving the python server running?

@dmonagle I just hit CMD + R in Xcode 🙂 and it does the job - site is regenerated. Then I switch to Chrome to refresh it. I also use `Network > Disable Cache` option in Chrome's console so everything is up to date just after refresh.

dmonagle:
Cheers @ncreated, I've spent so much time keeping the console and the Xcode environment separate in my mind due to them using different build directories that I never even thought about the fact that this tool will use a common Output directory.

JohnSundell:
I've now pushed my fix for the issue causing the `localhost` server to not be properly shut down here: https://github.com/JohnSundell/Publish/pull/27.

I also use the same workflow as @ncreated. I just leave the localhost server running in a separate terminal window all day, and generate my site through Xcode 👍

dmonagle:
All good. I'll close this request.

JohnSundell:
Cool, thanks a lot anyway @dmonagle. Really appreciate you contributing to Publish 🙂. And sorry if it took me a while to respond to this, but felt like I had to address a few other issues and PRs first.

Regarding your earlier questions:

- I've got a hot reloading tool that automatically reloads my web browser every time that I make a change to my site. I'll open source that very soon, I think it should give a lot of people in the community a big productivity boost 🚀. Just need to figure out exactly how to package it.
- I wouldn't want to bring Vapor in as a dependency, it's an amazing framework, but it feels a bit too heavy to be used as a simple `localhost` server for static files. One thing I would be very interested in seeing is if we could build a simple local web server using `SwiftNIO`. If it turns out to be quite simple, then that'd be a great alternative to the current Python-based server. Another option to explore would be to use Foundation + Cocoa to build a socket-based server or something.

javb99:
I've been messing around with using a modified version of this Swift NIO example server. It seems to work pretty well. https://github.com/apple/swift-nio/tree/master/Sources/NIOHTTP1Server

JohnSundell:
@javb99 Very interesting 🙂 That specific example contains a bit more code than I think I'd be willing to maintain for this type of feature, but perhaps we could simplify it to be specific for serving only `localhost` content? If you want to submit any implementation that you have as a WIP PR I'd love to discuss it further 👍

insidegui:
The [Network](https://developer.apple.com/documentation/network) framework would be an option, it's only available in Darwin tho :/

JohnSundell:
Using a Darwin-only solution is completely fine I think. I suspect most people will do their Publish-based web development on macOS anyway, at least for now, plus we could always fall back to the current Python-based solution on other operating systems for maximum compatibility 🙂

peteschaffner:
> * I've got a hot reloading tool that automatically reloads my web browser every time that I make a change to my site. I'll open source that very soon, I think it should give a lot of people in the community a big productivity boost 🚀. Just need to figure out exactly how to package it.

@JohnSundell I'm very intrigued about your hot reloading tool. Any idea on when that might be made available?

ccjensen:
checking out publish, but am finding that some sort of hot-reloading would be useful. Can't find any further references to the release of this reloading tool; did it ever get released?

PR #9, opened 2020-01-01
Keep camel case in siteName when generating a project

wacumov:
This PR fixes the following issue:
For example, we have a folder named `DeliciousRecipes` and we run `publish new` there. 
The website name in the generated files will be `Deliciousrecipes`, not `DeliciousRecipes `.


Tests/PublishTests/Tests/ProjectGeneratorTests.swift
```diff
@@ -0,0 +1,43 @@
...
+    func testMakeSiteNameFromLowercasedFolderName() {
+        XCTAssertEqual(ProjectGenerator.makeSiteName(folderName: "name"), "Name")
+    }
+    
+    func testMakeSiteNameFromCamelCaseFolderName() {
+        XCTAssertEqual(ProjectGenerator.makeSiteName(folderName: "CamelCaseName"), "CamelCaseName")
+    }
+    
+    func testMakeSiteNameFromFolderNameWithNonLetters() {
+        XCTAssertEqual(ProjectGenerator.makeSiteName(folderName: "My website 1"), "Mywebsite")
```
nicholascross: Is see this behaviour was in the original implementation as well, is it required?  Could the folder name be used as is - perhaps with quotes escaped?

folder name: `my "COOL" WeBsIt3` -> site_name `my &quot;COOL&quot; WeBsIt3`

Sorry if I have missed the original intent of including only letters but it seems like it could be valid to have other characters here.

> og:site_name - If your object is part of a larger web site, the name which should be displayed for the overall site. e.g., "IMDb".
https://ogp.me/

wacumov: Swift struct names can't start with digits and can't contain spaces. 
Technically the generated name can contain non-letters, but I think that this requirement (site names shouldn't contain non-letters) helps to keep things simple.

Sources/PublishCLICore/ProjectGenerator.swift
```diff
@@ -19,7 +19,7 @@ internal struct ProjectGenerator {
         self.folder = folder
         self.publishRepositoryURL = publishRepositoryURL
         self.publishVersion = publishVersion
-        self.siteName = String(folder.name.capitalized.filter { $0.isLetter })
+        self.siteName = Self.makeSiteName(folderName: folder.name)
```
JohnSundell: I think this could be implemented in a slightly simpler way, by extending `String` with a `asSiteName()` method that does the following:
```swift
private extension String {
    func asSiteName() -> Self {
        let name = prefix(1).capitalized + dropFirst()
        return name.filter { $0.isLetter }
    }
}
```
Then the assignment here could simply be: `self.siteName = folder.name.asSiteName()`.

Tests/PublishTests/Tests/ProjectGeneratorTests.swift
JohnSundell: Also keep in mind that this is just for the `publish new` convenience command. A user can easily customize the generated project to include numbers, if they want.

Tests/PublishTests/Tests/ProjectGeneratorTests.swift
```diff
@@ -0,0 +1,43 @@
+/**
+*  Publish
+*  Copyright (c) John Sundell 2019
+*  MIT license, see LICENSE file for details
+*/
+
+import XCTest
+@testable import PublishCLICore
```
JohnSundell: I prefer not using `@testable import` within frameworks, since it often leads to testing too many implementation details, rather than the public API. There's already a test case called `CLITests` where you can perform end-to-end testing of your new functionality - which I think is the proper way to test CLI tools, that they actually produce the right outcome.

Sources/PublishCLICore/ProjectGenerator.swift
wacumov: Yes, this implementation is better. I use it in the second commit. But I slightly changed your variant to handle cases like "1-website".

Tests/PublishTests/Tests/ProjectGeneratorTests.swift
wacumov: I moved all the tests related to siteName to CLITests and got rid of `@testable import`.

PR #10, opened 2020-01-01
PublishCLI apply easy update by `publish update`

ytyubox:
`publish update` to clone/pull the Publish

help:
1. No idea how to test it
2. $TMPDIR maybe not a good idea


JohnSundell:
Thanks a lot for this @ytyubox, but I don't think Publish should take on the responsibility of updating itself. Such functionality is better left to package managers, or other kinds of tools that are dedicated to the task of managing command line tools. For now, I think the best is to update the Publish CLI manually, but I might look into adding Homebrew support in the future. Closing this, but thanks anyway.

PR #13, opened 2020-01-02
Allow excluding section's path as well while generating SiteMap

artrmz:
Use case:
I want to have a whole section which I do not want to be included in the SiteMap.

### DIFF
--- a/Tests/PublishTests/Tests/SiteMapGenerationTests.swift
+++ b/Tests/PublishTests/Tests/SiteMapGenerationTests.swift
@@ -60,6 +60,7 @@ final class SiteMapGenerationTests: PublishTestCase {
 
         let unexpectedLocations = [
             "https://swiftbysundell.com/one/itemB",
+            "https://swiftbysundell.com/two",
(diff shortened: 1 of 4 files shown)


artrmz:
Added tests.

Tests/PublishTests/Tests/SiteMapGenerationTests.swift
```diff
@@ -57,6 +60,8 @@ final class SiteMapGenerationTests: PublishTestCase {
 
         let unexpectedLocations = [
             "https://swiftbysundell.com/one/itemB",
+            "https://swiftbysundell.com/two/itemC",
```
JohnSundell: I'd also add the section's index location here as well:
```
https://swiftbysundell.com/two
```
Just to protect against future regressions. I think it's also worth documenting that adding a section's path to the list of excluded paths removes that entire section (including its items).

artrmz: Sounds good 👍

PR #14, opened 2020-01-02
Use @autoclosure on PublishingStep.if

javb99:
Don't evaluate the `step` if the condition is `false`. I believe this is more inline with Swift's short-circuit conditional behavior.


javb99:
That is a great point. I hadn’t considered that. Thanks for the reply.

JohnSundell:
No worries. It's kind of unfortunate that we have to design our APIs around Xcode/SourceKit's limitations, but that's the way things are at the moment. Will close this for now, but I'd love to revisit it once the autocomplete for `@autoclosure` arguments have been improved 👍

PR #15, opened 2020-01-03
Allow plugins to be added to default pipeline

slashmo:
When adding the Splash plugin to my deployment pipeline I noticed that the only thing I wanted to change to the default pipeline was installing the plugin. I tried installing it through the `additionalSteps` API but they are called too late so they don't come into play. 

I wasn't sure whether to change the implementation to call those steps prior to anything else so I added an additional `plugins` parameter. 

Also, I wasn't sure how to add proper unit tests for this because the default pipeline is currently not tested at all, but I'd be glad to look into this more.

### DIFF
--- a/Sources/Publish/API/Website.swift
+++ b/Sources/Publish/API/Website.swift
@@ -62,7 +62,7 @@ public extension Website {
     /// - parameter deploymentMethod: How to deploy the website.
     /// - parameter additionalSteps: Any additional steps to add to the publishing
     ///   pipeline. Will be executed right before the HTML generation process begins.
-    /// - parameter plugins: Plugins to be installed beforehand.
+    /// - parameter plugins: Plugins to be installed at the start of the publishing process.
     /// - parameter file: The file that this method is called from (auto-inserted).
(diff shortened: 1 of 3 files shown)


Sources/Publish/API/Website.swift
```diff
@@ -62,6 +62,7 @@ public extension Website {
     /// - parameter deploymentMethod: How to deploy the website.
     /// - parameter additionalSteps: Any additional steps to add to the publishing
     ///   pipeline. Will be executed right before the HTML generation process begins.
+    /// - parameter plugins: Plugins to be installed beforehand.
```
JohnSundell: Very nitpicky, but perhaps specifying "beforehand" a bit more to say "at the start of the publishing process"?

slashmo:
Ah, somehow the Xcode test coverage report didn't include the `publish(withTheme:)` method, but I checked again and it got reported correctly. Thanks for the hint 👍

kimar:
Awesome! I was actually coming up with a whole pipeline just because I needed to hook Splash in before the Markdown is parsed. This PR looks like a pretty nice fix for this situation.

PR #19, opened 2020-01-04
Apply new content, for publish new [FilePath]

ytyubox:
I found `publish init` is more likely to express create a brand new website, Just replace `publish new` with creating content at the path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,12 +285,31 @@ $ cd Publish
 $ make
 ```
 
-Then, create a new folder for your new website project and simply run `publish new` within it to get started:
+Then, create a new folder for your new website project and simply run `publish init` within it to get started:
 
 ```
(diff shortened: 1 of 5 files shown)


JohnSundell:
Thanks for this @ytyubox, but I don't think it's a good idea to rename `publish new`, and I also don't wish to add this sort of file generation functionality to the command line tool at the moment. Publish was built to be very dynamic, so what kind of metadata that each site's items will require will likely differ quite a lot - to the point where I think it's better for each developer to have their own set of templates that they manage in order to create new content. The command line tool is also meant to be kept really simple, so I'd ideally like it to not gain any significant new features at the moment.

JohnSundell:
Closing this one. Thanks again @ytyubox, and sorry that this didn't make it in.

PR #20, opened 2020-01-04
Enable dynamic port configuration for run CLI command

jlorfeo:
On my mac the port 8000 was blocked, therefore I extended the run CLI command to support a dynamic config with the attributes "-p" and "--port".
Because the CLI seems to be lightweight I only introduced a command option for this specific usecase, but with enums etc. there could be the posibility of generally introducing options for existing CLI commands.

### DIFF
--- a/Sources/PublishCLICore/CLI.swift
+++ b/Sources/PublishCLICore/CLI.swift
@@ -1,11 +1,11 @@
 /**
-*  Publish
-*  Copyright (c) John Sundell 2019
-*  MIT license, see LICENSE file for details
-*/
+ *  Publish
+ *  Copyright (c) John Sundell 2019
+ *  MIT license, see LICENSE file for details
+ */
 
-import Foundation
 import Files
+import Foundation
 import ShellOut
 
 public struct CLI {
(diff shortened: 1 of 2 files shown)


alexito4:
> On my mac the port 8000 was blocked

Are you sure it wasn't blocked by a previous run of publish CLI? I've noticed that the python server doesn't get kill when the CLI stops, so trying to run it again always fails until you kill it.

jlorfeo:
Thanks for your suggestion! :)
I had another application running on that port but I will keep the issue in mind.

Sources/PublishCLICore/CLI.swift
```diff
@@ -42,8 +42,13 @@ public struct CLI {
             let deployer = WebsiteDeployer(folder: folder)
             try deployer.deploy()
         case "run":
-            let runner = WebsiteRunner(folder: folder)
-            try runner.run()
+            if arguments.count > 2, arguments[2].contains("-p=") || arguments[2].contains("--port="), let portNumber = extractPortNumber(from: arguments[2]) {
```
JohnSundell: This statement is quite hard to follow. Mind separating it into a dedicated method instead? I'm thinking the call site could look something like this:
```swift
let portNumber = extractPortNumber(from: arguments)
let runner = WebsiteRunner(folder: folder, portNumber: portNumber)
try runner.run()
```
That way the `extractPortNumber` method could contain all of this complexity, and you could use the `guard` statement to exit early, keeping the logic very neat and clean.

Sources/PublishCLICore/CLI.swift
```diff
@@ -63,10 +68,17 @@ private extension CLI {
 
         - new: Set up a new website in the current folder.
         - generate: Generate the website in the current folder.
-        - run: Generate and run a localhost server on port 8000
-               for the website in the current folder.
+        - run: Generate and run a localhost server on default port 8000
+               for the website in the current folder. Use the
```
JohnSundell: Think you could move the "-p" up here to make the text wrap a bit more neatly. Nitpick, I know 😅

jlorfeo: I understand that ;) I fixed it and it does look nicer ;)

Sources/PublishCLICore/CLI.swift
jlorfeo: I saw the "=" notation on command lines quite often, therefore I went with that.
Your approach reduces complexity, therefore I changed the "=" :). Also I caught myself multiple times writing "-p 3000" :D.
The complete logic of validating and returning the port number from the cli arguments was moved in the "extractPortNumber" method

PR #21, opened 2020-01-04
Proposal to support Slug for Cool URI support

steve-h:
The current Publish implementation uses the filename as the Cool URI. It seems to be determined before the Markdown front-matter is parsed and decoded. The publishing path should be determined after parsing the front-matter.  I believe the inclusion in a Tag Index page occurs after such front-matter declares the tags.

See the FileOrganization.md for some reference links.

### DIFF
--- a/FileOrganization.md
+++ b/FileOrganization.md
@@ -1,0 +1,11 @@
+## Cool URL management
+
+[Cool URIs don't change](https://www.w3.org/Provider/Style/URI) implies that the best practice is to keep a nice URI for your content.
+
+[Clean URL](https://en.wikipedia.org/wiki/Clean_URL#Slug) shows how a Slug feature can help with the cool URI.  
+
+The slug offers an alternative to using the markdown filename as the URI ending.  Many implementations use the Markdown frontmatter to specify the slug. (such as [Hugo Slug](https://gohugo.io/content-management/organization/#slug-1) )
+
+The current Publish implementation uses the filename as the Cool URI. It seems to be determined before the Markdown front-matter is parsed and decoded. The publishing path should be determined after parsing the front-matter.  I believe the inclusion in a Tag Index page occurs after such front-matter declares the tags.
+
+At the moment I am publishing this discussion paper, rather than submitting code as the code change looks significant.
--- a/Package.resolved
+++ b/Package.resolved
@@ -15,8 +15,8 @@
         "repositoryURL": "https://github.com/johnsundell/files.git",
         "state": {
           "branch": null,
-          "revision": "941052daf1c0aa6999343ee899dd3d743c86f211",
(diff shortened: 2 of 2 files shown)


JohnSundell:
Thanks for this suggestion @steve-h. I'm curious what kind of concrete use case this type of functionality is needed for?

steve-h:
I guess a concrete case is moving my Hugo site. I have to change all the filenames to keep my cool URI paths the same when moving my site.  There are many Hugo users that will need a tool to look at the front-matter slug and rename their markdown files. Just a real-life markdown portability argument.

I guess you had all your filenames coded to get nice URLs like https://www.swiftbysundell.com/articles/the-decade-of-swift/

Some multilingual backend CMS systems may use a different file naming than the nice URI?

JohnSundell:
I see. I'd be happy to let an item's path be optionally determined by a `path:` metadata value, rather than its file path, just like how an item's title can be overridden using `title:`. Should be a really small change. Would that solve the issue?

JohnSundell:
To answer your question, yes, my file names all match their URLs. So the Markdown file for `/articles/the-decade-of-swift` is located at `Content/articles/the-decade-of-swift.md`. I personally wouldn't want it any other way, I think that'd be awfully messy. But again, happy to add that `path` override if it makes it easier to use Publish for others 🙂

steve-h:
I think [slug](https://en.wikipedia.org/wiki/Slug_(publishing)) is the historical publishing industry name for a short code name of a story/article and it looks like Wordpress, Django and Hugo use slug. Path is the URI implementation of the delivery of the slug in a particular web publishing role. I suggested using slug from the Hugo perspective as I would still have to visit all the files if you pick a different meta name.

slug was a chunk of lead in printing press; no one told them about path back then. The name persists in the domain and I see no reason to pick a different name for it. Especially when it is just the last piece of the path and not the whole path.

JohnSundell:
I don't mean for this to sound arrogant or dismissive, but at the moment I'm much more concerned with keeping Publish internally consistent than making it consistent with other static site generators. Since this piece of data is known all throughout Publish as `path`, I'd like to keep that naming in the Markdown metadata as well, at least for now.

I'm also hoping to make `Item.path` a mutable property rather soon, so that anyone who wants to use `slug` (or any other name) could simply add that to their own `ItemMetadata` and write a quick plugin that re-assigns each item's path accordingly.

The idea is that wherever I've made strong decisions around content structure or metadata when designing Publish, you should be able to override that using either custom metadata, plugins, or configuration options. We're not 100% there yet, so feedback like this definitely helps a lot.

steve-h:
Ok. I will just change my file names for now and leave my slug data alone. That way I am in the same boat as your sites. No change needed on this PR.

PR #23, opened 2020-01-05
Adding support for absolute path strings

fishcharlie:
Currently when calling the `head(for: , on: )` function and passing in a `stylesheetPaths` array, there is no way to pass in absolute paths. Which makes it impossible to import external stylesheets into the site without using the alternate `head` function and having to specify every single option for head. Which makes the `head(for: , on: )` function much less useful for the majority of use cases.

The goal of this PR is to fix that by adding an `absolute` boolean to the `Path` struct, and returning the string for absoluteString if that boolean is set to true.

This is really meant to be an initial PR. There are probably better ways to do this, but I don't quite have enough knowledge about the codebase to know what those are at this point. So feedback is welcome!

### DIFF
--- a/Sources/Publish/API/Path.swift
+++ b/Sources/Publish/API/Path.swift
@@ -10,6 +10,7 @@ import Foundation
 /// location or to a resource, such as a file or image.
 public struct Path: StringWrapper {
     public var string: String
+    public var absolute: Bool
 
     public init(_ string: String, absolute: Bool = false) {
         self.string = string
--- a/Sources/Publish/API/Path.swift
+++ b/Sources/Publish/API/Path.swift
@@ -11,8 +11,9 @@ import Foundation
 public struct Path: StringWrapper {
     public var string: String
 
-    public init(_ string: String) {
+    public init(_ string: String, absolute: Bool = false) {
         self.string = string
+        self.absolute = absolute
     }
 }
 
@@ -27,7 +28,7 @@ public extension Path {
     /// Convert this path into an absolute string, which can be used to
     /// refer to locations and resources based on the root of a website.
     var absoluteString: String {
-        if string.first == "/" { return string }
+        if string.first == "/" || self.absolute { return string }
         return "/" + string
     }
 


fishcharlie:
It is super late here. So I'm unable to look at the test failures right now. If anyone gets a chance to before me and can figure out what is going on here, it'd be awesome if you could leave a comment explaining your findings!

I have enabled `Allow edits from maintainers.`, so @JohnSundell (I don't think there are any other maintainers at this point) should be able to push new changes to this PR (if you want to make changes or anything).

Sources/Publish/API/Path.swift
```diff
@@ -11,8 +11,9 @@ import Foundation
 public struct Path: StringWrapper {
     public var string: String
 
-    public init(_ string: String) {
+    public init(_ string: String, absolute: Bool = false) {
```
jlorfeo: Your problem seems to be that there is no property declaration for "absolute" . Its only mentioned in the constructor

fishcharlie: @jlorfeo Right... I should have caught that. Downsides to coding when tired haha. Thanks so much!

JohnSundell:
The problem is not a test failure. The project fails to build because with this change, `Path` no longer conforms to `StringWrapper`. Simply building the project in Xcode should let you reproduce the problem.

I definitely think that this should be fixed, but since the rules around what makes an absolute path (or URL) are quite straightforward (AFAIK), I think it makes more sense for Publish to take on deciding how to render each path (just like how it does with absolute paths within the same website). So when returning a value from `absoluteString`, I'd just check if the path has `http://` or `https://` as a prefix, and if so return that string unmodified, rather than requiring the user to pass an additional argument.

fishcharlie:
@JohnSundell Makes sense. By test failure I more meant CI failure haha. I made this change in the GitHub interface, so definitely didn't take as much time as I should have on it.

That would be a solution for sure, and something that crossed my mind. The only problem I have with it, is technically the protocol can be anything (ex. `ftp`). Now, that might not be an issue for this project, due to the scope of it primarily being a system for generating static web content. But it is an edge case that should at least be thought of. I'm not opposed to it at all, but I don't know if there are plans to use the `Path` struct in places where having support for additional protocols is necessary. Having the API be opinionated here might be the right decision, but it wasn't something I immediately wanted to jump to right off the bat.

Any thoughts about how to proceed here? Would you like me to make the change to check for the `http://` or `https://` prefix and return the string unmodified if so? Would you like me to close this PR and you (or I) can create a new one with that work?

One other option is to do a regex, that checks for 1 or more characters followed by `://` then followed by at least 1 more character. Not sure if there would be false positives with that, but it would handle protocols (other than http & https) better. Something like `/.+:\/\/.+/` maybe? Might be overkill for this use tho.

fishcharlie:
@JohnSundell Any updates or thoughts on this? Really wanna use this for a project I’m working on, but this is a major blocker, and I’m not sure how to work around it without doing a really ugly temporary solution that will need to be changed after this gets fixed. Which doesn’t seem like it’d be worth the level of effort required.

JohnSundell:
@fishcharlie If you're blocked, please fork Publish and implement a temporary fix until a proper one has been merged into master (for example the one you submitted as part of this PR), and depend on your fork for now. Will take a look at this as soon as I can.

JohnSundell:
Fixed here: https://github.com/JohnSundell/Publish/pull/33. Went with the simplest solution of whitelisting `http://` and `https://`, for now. Thanks for reporting this issue, @fishcharlie 👍

fishcharlie:
@JohnSundell Perfect! Any sense of an ETA for the next release that this will be included in? I haven't gotten around to using a fork with a fix, so might just wait for a release here haha. Thank you so much!!!

JohnSundell:
My pleasure, it was a great find - I'm sure other people have/would have encountered this issue as well 👍

I'll most likely do another release later today, but you never have to wait for maintainers to release a new version to use the latest patches. Just depend on `master` (when the PR is merged in) directly and you'll get it - using this syntax within your `Package.swift`'s `dependencies`:

```
.(url: "https://github.com/johnsundell/publish.git", .branch("master"))
```

That's the beauty of open source - you are in control. You don't have to wait for a new "product" to be officially released, although once a new release is made available I'd recommend switching over to it, for safety 🙂

PR #25, opened 2020-01-07
Closes down SimpleHTTPServer after spawn

davidskeck:
The SimpleHTTPServer continues to run in the background normally meaning you can't have successive calls to ```publish run``` without an error occurring. This change kills the server task correctly after use.

ShellOut will throw an error, but the task is killed. Example run:

```
Press any key to stop the server and exit
 
🔌 Shutting down...
Encountered error: 
❌ ShellOut encountered an error
Status code: 9
Message: ""
Output: ""
```

I'm open to suggestions on how to do this another way, but wanted to get this out here in case others were facing the same issue as me when trying to iterate.

Reference for shell command: https://stackoverflow.com/questions/12647196/how-do-i-shut-down-a-python-simplehttpserver

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -37,6 +37,14 @@ internal struct WebsiteRunner {
         }
 
         _ = readLine()
+        
+        print("""
+        🔌 Shutting down...
+        """)
+        _ = try shellOut(
+            to: "kill -9 `ps -ef | grep SimpleHTTPServer | grep \(self.portNumber) | awk '{print $2}'`",
+            at: outputFolder.path
+        )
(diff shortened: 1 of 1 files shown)


JohnSundell:
Thanks for this @davidskeck, but I think this solution is quite dangerous. Since it uses the `kill` command, a slight bug in this code could end up terminating the wrong process - which could lead to data loss. I'm developing a proper fix for this issue, which involves closing the underlying `Pipe` when `publish run` exits. It should be ready within a day or two, so closing this PR.

davidskeck:
Sounds good! Looking forward to seeing your implementation.

JohnSundell:
Here it is 😀 https://github.com/JohnSundell/Publish/pull/27

PR #27, opened 2020-01-07
CLI: Stop local web server when ‘publish run’ exits

JohnSundell:
This patch fixes a CLI bug that would cause the local web server started by `publish run` to keep running indefinitely. Now, the server is properly shut down whenever the CLI exits.

Error handling has also been improved, and in case there’s already a `localhost` server running on the same port, a much richer error message with suggestions will now be displayed:

```
❌ Failed to start local web server:
A localhost server is already running on port number 8000.
- Perhaps another 'publish run' session is running?
- Publish uses Python's SimpleHTTPServer, so to find any
  running processes, you can use either Activity Monitor
  or the 'ps' command and search for 'python'. You can then
  terminate any previous process in order to start a new one.
```

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -42,8 +42,8 @@
         "repositoryURL": "https://github.com/johnsundell/shellout.git",
         "state": {
           "branch": null,
-          "revision": "4ebf25863deb9c3c02696704fc3d39736183f258",
-          "version": "2.2.1"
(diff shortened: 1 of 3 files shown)


davidskeck:
I was curious how you'd hold onto the process to terminate it correctly. Nice work!

JohnSundell:
Thanks, @davidskeck! 🙂

PR #29, opened 2020-01-08
Item resource filename

dmonagle:
This PR is reliant on the acceptance of #28 as it makes little sense without it.

Consider the following files making up the content for a section of a site:

- 01_installation
- 02_getting-started
- 03_advanced-techniqes

In this hypothetical site, it is important to present these items in order in the section. I have personally run into this when wanting to order sections containing a list of features and FAQs. I was solving this with an Item metadata field called **position**.

The issue with using the position metadata is that the filesystem will not sort based on this and it becomes a management headache if you want to be able to review site content in the correct order.

With the ability to access item.resourceFilename items can be sorted thus:

```swift
items.sorted(by: { $0.resourceFilename ?? "" < $1.resourceFilename ?? "" })
``` 

Then in the content of each file, the metadata for **path** can be set to the appropriate path for the file.

Neither this PR or #28 changes any fundamental way Publish operates and have no backward compatibility issues.

### DIFF
--- a/Sources/Publish/API/Item.swift
+++ b/Sources/Publish/API/Item.swift
@@ -19,7 +19,10 @@ public struct Item<Site: Website>: AnyItem, Hashable {
     /// this item is for.
     public var metadata: Site.ItemMetadata
     public var tags: [Tag]
+    /// The absolute path to this item relative to the root of the site
     public var path: Path { makeAbsolutePath() }
+    /// The name of the resource file used to generate this item
+    public let resourceFilename: String?
     public var content: Content
     public var rssProperties: ItemRSSProperties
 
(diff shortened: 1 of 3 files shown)


JohnSundell:
Thanks for this @dmonagle. I definitely think this is a use case that should be supported, but I'm thinking that there might be a more "elegant" solution here that doesn't involve adding a new property. How about we ensure that all Markdown files are added in alphabetical order by default? That way, you shouldn't need to perform any additional sorting or filtering on them. What do you think?

dmonagle:
I absolutely agree with you. The less properties added the better. I only resisted the suggestion of changing the default sort ordering as it is effectively a breaking change that might surprise people. I do, however, think it’s the best way to go as sorting by date is already easily achieved. I’ll do this and open another pull request.

JohnSundell:
That sounds great. I also agree that keeping breaking changes to a minimum is really important, but I think it's fine to modify things to get them right before we hit `1.0.0` (after which things will be more set in stone, at least until the next major release). I also think that having files be parsed in alphabetical order is what most people would expect, and no-one should be making assumptions about the item order anyway, since it was essentially undefined at this point.

JohnSundell:
Can this be closed?

dmonagle:
Sure. This solution won’t be merged regardless so any other way to solve the problem can be made as a separate PR.

JohnSundell:
Sounds great, thanks again @dmonagle!

PR #32, opened 2020-01-08
Upgrade http server to Python 3

manuzhang:
Python 2 has been [sunset since Jan 1st, 2020](https://www.python.org/doc/sunset-python-2/) and [`SimpleHTTPServer` was merged into `http.server`  in Python 3](https://docs.python.org/2/library/simplehttpserver.html)

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -30,7 +30,7 @@ internal struct WebsiteRunner {
         serverQueue.async {
             do {
                 _ = try shellOut(
-                    to: "python -m SimpleHTTPServer \(self.portNumber)",
+                    to: "python -m http.server \(self.portNumber)",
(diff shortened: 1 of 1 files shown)


JohnSundell:
Thanks for this @manuzhang, but since macOS doesn't come with Python 3 pre-installed, and since `SimpleHTTPServer` works great even though it's using Python 2, I don't see any reason to force Publish users to install a separate version of Python just to be able to run a `localhost` server. We're also looking to move to a Swift solution for the local server soon. Closing this, but thanks anyway.

manuzhang:
@JohnSundell Make sense. Not depending on Python will be best. Would it be valuable to add a note to README such that people with Python 3 could work around with one line change ?

JohnSundell:
@manuzhang Even better, I think, would be to do this automatically. Run `python --version`, check if the user is running Python 3, and if so use the new syntax. That's a change I'd love to see 🙂

PR #38, opened 2020-01-10
Fixed SectionID rawvalue bug in item.

Goosse:
When making a multi-word SectionID it is necessary to set the rawvalue explicitly to for more standard urls.
Doing this caused two folders to be created in Output. If the SectionID was `case exampleSection = "example-section"`, Output would contain `example-section/index.hml` and `exampleSection/actualContents-of-folder-in-Content`

### DIFF
--- a/Sources/Publish/API/Item.swift
+++ b/Sources/Publish/API/Item.swift
@@ -56,6 +56,6 @@ internal extension Item {
 
 private extension Item {
     func makeAbsolutePath() -> Path {
-        "\(sectionID)/\(relativePath)"
+        "\(sectionID.rawValue)/\(relativePath)"
     }
 }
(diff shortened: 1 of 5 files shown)


JohnSundell:
That's an excellent find, and thanks for submitting a fix for it @Goosse. Do you mind also implementing a simple test for it, to make sure that this type of issue doesn't happen again? Let me know if you'd like any guidance on how to implement such a test.

Goosse:
Sure, but tests are really uncharted territory for me yet.  Would appreciate some help.

Goosse:
I poked around a little in the tests and think I came up with something that tests it appropriately.
Specifically `testGeneratingStandAloneHTMLFiles` will fail if the previous commit is undone and passes when the bug is fixed.  Like I said, I don't have much practice with tests, so please advise if something could be done in a better way.

PR #40, opened 2020-01-10
Mention front matter

alexito4:
The "header" in a markdown file is commonly known as Front Matter in many other static site generators. For people coming from other generators it would be useful to relate what Publish calls "metadata" to what they know as "front matter". It would also help for people that does a search about it.

I would also like to include a mention of this in the README but I don't think it fits anywhere in the current prose. 

Example of other well known generators using this wording:
- Gatsby https://www.gatsbyjs.org/docs/adding-markdown-pages/#frontmatter-for-metadata-in-markdown-files
- Jekyll https://jekyllrb.com/docs/front-matter/
- Hugo https://gohugo.io/content-management/front-matter/

### DIFF
--- a/Documentation/HowTo/custom-markdown-metadata-values.md
+++ b/Documentation/HowTo/custom-markdown-metadata-values.md
@@ -1,6 +1,6 @@
 # How to express custom metadata values using Markdown
 
-Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file (commonly known as *front matter*).
+Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file (within other tools referred to as *front matter*).
 
 Let’s say that we’re building an shopping website, and that we’ve defined a custom `productPrice` item metadata value, like this:
 
--- a/Documentation/HowTo/custom-markdown-metadata-values.md
+++ b/Documentation/HowTo/custom-markdown-metadata-values.md
@@ -1,6 +1,6 @@
 # How to express custom metadata values using Markdown
 
-Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file.
+Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file (commonly known as *front matter*).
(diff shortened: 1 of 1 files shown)


Documentation/HowTo/custom-markdown-metadata-values.md
```diff
@@ -1,6 +1,6 @@
 # How to express custom metadata values using Markdown
 
-Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file.
+Publish enables each website to define its own site-specific item metadata, through its `Website.ItemMetadata` type. When adding items using Markdown, those values can then be expressed by adding a metadata header at the top of a file (commonly known as *front matter*).
```
JohnSundell: I wouldn't go so far as to say that "front matter" is "commonly known" (I only heard about it for the very first time the other day myself). How about phrasing this sentence as "within other tools referred to as *front matter*"?

alexito4: that sounds good. let me do the change 😉

alexito4:
For the curious readers > https://www.scribendi.com/advice/front_matter.en.html

Goosse:
Another thought, related to this. (Didn't think this was worth a separate pull request just hear what you think.)  It seems that most static site generators use YAML in their metadata/front matter.  All 3 linked above do.  If Publish does not do this, it can cause issues when integrating other tools that assume metadata will be in YAML.  

Example: https://www.netlifycms.org
Netlify CMS will include a string in single quotes.  So tags would look like this:
`tags: 'test, one, two, three'`
Currently Publish decodes this as `["'test", "one", "two", "three'"]`
It includes the single quotes in the first and last tags.  This seems counter intuitive.  

Any thoughts?

JohnSundell:
I personally don't think using a complete YAML implementation for Markdown metadata is a good idea, as I think it adds a ton of complexity for very little benefit (in fact, I think the way Publish is currently designed encourages users to keep their metadata simple, which will most likely be beneficial in the long run). I think compatibility with external systems (such as Netlify) should be achieved using plugins (and if that's not possible for some reason, we should make it possible, rather than modifying Publish's own data design).

Goosse:
I agree. I'm going to try working on a YAML decoder plugin.

Goosse:
Ok, I started poking around to understand how plugins work and how this would be accomplished. It would seem I would need create a Modifier (declared in Ink), but as I understand it, there is no option for a metadata target.  Am I at least looking in the right direction?

JohnSundell:
I just defined such modifier targets that you should be able to use to achieve what you're looking for 🙂 https://github.com/JohnSundell/Ink/pull/39

PR #41, opened 2020-01-10
Make CommandLine extension public

alexito4:
One really nice thing about Publish is its command line output, showing each step of the pipeline and with fancy emojis:

```
Publishing Multiverse (6 steps)
[1/6] Copy 'Resources' files
[2/6] Add Markdown files from 'Content' folder
[3/6] Sort items
[4/6] Generate HTML
[5/6] Generate RSS feed
[6/6] Generate site map
✅ Successfully published Multiverse
```

But currently there is no way to reuse the code for this in our own Steps or Plugins, which makes it harder to add our own output in a way that maintains the style of Publish. 

I think that for Plugins this is crucial since it would be really nice to be able to integrate fully with Publish and offer the best end user experience. In my case I would like to display a nice error, see https://github.com/alexito4/ReadingTimePublishPlugin/pull/2.

This is a Draft, the only thing I've done is making the extension public, in order to get your views on this topic. If you're cool with it I can proceed by moving the file to the API folder and anything else you consider necessary (tests? docs?)

### DIFF
--- a/Sources/Publish/Internal/CommandLine+Output.swift
+++ b/Sources/Publish/Internal/CommandLine+Output.swift
@@ -6,7 +6,7 @@
 
 import Foundation
 
-internal extension CommandLine {
+public extension CommandLine {
     enum OutputKind {
         case info
         case warning


JohnSundell:
@alexito4, thanks for this, but I don't really understand why it's necessary to make this public in order for you to add the output or logging you're looking for. If you look at the implementation of this utility, it's just a simple wrapper around `fputs`. Why not simply call that (or `print()`) directly, why do you need this wrapper? It won't give you the step prefixes, and only adds an emoji in case of success (which is super trivial to replicate in your own code). Publish already facilities outputting errors in a nice manner, even if they're thrown from plugins. Just throw an error from within your plugin implementation and it should all be handled (and if not, we should improve that, rather than moving this implementation detail to the public API).

alexito4:
You're right that we can just print and this doesn't add the step number, but it does add the different emojis based on the `OutputKind`. 

You're also right in that Publish reports errors that are thrown, but that doesn't help if what a Plugin wants to report is a warning or just a success. 

In my mind it was a good idea to let the plugins share some infrastructure with the main framework and also let them have the same nice output. But I would accept that maybe prefixing emojis is not worth the API surface area.

JohnSundell:
Yeah, I personally don't think it's worth it right now. If we end up adding some more sophisticated logging/warning reporting in the future (which is why that unused `warning` case is there actually, since I ended up cutting such a system in the end, that I had originally intended) then I'd be happy to discuss this again, but for now I think we can let each plugin do their own logging and see how things play out. Sounds fair? 🙂

alexito4:
Iep makes sense. Thanks for considering. I will close this 😉

PR #44, opened 2020-01-12
Handle capitalised default metadata keys

bellebethcooper:
This PR is designed to support metadata keys in Markdown files that start with a capital letter, rather than being all lowercase. Coming from Pelican, all my files have capitalised metadata keys, and if we can support those, anyone switching from a similar setup won't need to convert their old files. It's still a WIP but I'd love to get some advice on how to move forward. Here's where I'm at:

1. I've created a `MetadataKeyDecodingStrategy` enum with values for `.lowercase` (default, all lowercase) and `.capitalized` (first letter of the key capitalised). The `PublishingContext` has `.lowercase` by default but this can be customised in a publishing step, just like setting a custom date formatter
2. I've written a test for handling capitalised _built-in_ metadata keys (i.e. date, description, audio, etc.) which is passing
3. I've created a struct to hold the built-in metadata keys so that a) the code base isn't full of hard-coded strings, and b) I can make that struct handle capitalising the underlying strings (or not) based on the `MetadataKeyDecodingStrategy` in use. I'm not 100% sure on this approach, and am open to suggestions, but the test is passing for this at least
4. I've written a test for handling capitalised _custom_ metadata keys, and this one is not passing yet. I'd love some advice on how to approach this
5. Part of handling custom capitalised keys was to change the way the `KeyMap` is created in `MarkdownMetadataDecoder`:

```swift
var key: Key?
switch keyDecodingStrategy {
    case .lowercase:
        key = Key(stringValue: stringKey)
    case .capitalized:
        key = Key(stringValue: stringKey.lowercased())
    }

if let key = key {
    evaluated.array.append(key)
    evaluated.set.insert(stringKey)
}
```

6. _But_ in my test, this code is not being hit, and the test is failing because this call: `let metadata = try Site.ItemMetadata(from: decoder)` inside `makeItem` uses the raw `Metadata` keys pulled directly from the Markdown. I'm wondering whether I might need to instead approach this by altering the actual `[String: String]` that's stored when parsing the Markdown, so that the raw keys are de-capitalised if needed before being stored, rather than trying to do this at a later step, such as in the code above where we're creating a `Key`. I _think_ if I stick with doing it at a later step, I'll need to keep the changes shown above, as well as adding changes to `ItemMetadata(from: decoder)`, but I may have misunderstood.
7. The last point to make is that this approach insists on the user knowing in advance which decoding strategy they want to use, and being consistent in just one strategy. I don't think this will be an issue, but it's worth mentioning. Another approach could be to always attempt to use the lowercase approach, and fall back to trying the capitalised (i.e. de-capitalising the raw keys from the Markdown) if the default approach doesn't work. This would mean the user doesn't have to set the key decoding strategy in advance, nor do they have to be consistent, but I think the code itself might be a little more messy that way, and I'm not sure it's necessary.

Hopefully that all makes sense! Very open to guidance on this.

### DIFF
--- a/Sources/Publish/API/PublishingContext.swift
+++ b/Sources/Publish/API/PublishingContext.swift
@@ -33,6 +33,8 @@ public struct PublishingContext<Site: Website> {
     public var allTags: Set<Tag> { tagCache.tags ?? gatherAllTags() }
     /// Any date when the website was last generated.
     public private(set) var lastGenerationDate: Date?
+    /// Determines how metadata keys should be decoded from Markdown files
+    public var keyDecodingStrategy: MetadataKeyDecodingStrategy = .lowercase
 
     private let folders: Folder.Group
     private var tagCache = TagCache()
@@ -278,7 +280,8 @@ internal extension PublishingContext {
     func makeMarkdownContentFactory() -> MarkdownContentFactory<Site> {
         MarkdownContentFactory(
             parser: markdownParser,
-            dateFormatter: dateFormatter
+            dateFormatter: dateFormatter,
+            keyDecodingStrategy: keyDecodingStrategy
         )
     }
 
--- a/Sources/Publish/Internal/MarkdownContentFactory.swift
+++ b/Sources/Publish/Internal/MarkdownContentFactory.swift
@@ -12,6 +12,7 @@ import Codextended
 internal struct MarkdownContentFactory<Site: Website> {
     let parser: MarkdownParser
     let dateFormatter: DateFormatter
+    let keyDecodingStrategy: MetadataKeyDecodingStrategy
 
     func makeContent(fromFile file: File) throws -> Content {
         let markdown = try parser.parse(file.readAsString())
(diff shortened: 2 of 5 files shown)


JohnSundell:
Thanks a lot for this @bellebethcooper, making the way Publish decodes Markdown metadata more flexible is a great goal, and something I definitely would like to support. After discussing this with you on Twitter, and seeing some other feedback related to Markdown metadata, I had an idea on how I think this could be implemented in a slightly more powerful and future-proof way.

I realized that we probably want this type of modification to happen at the Markdown parsing level, rather than within Publish, so I went ahead and implemented new modifier targets for `Ink` that should enable you to build a plugin that performs the capitalization that you're looking for. For example like this:

```swift
public extension Plugin {
    static func splash(withClassPrefix classPrefix: String) -> Self {
        Plugin(name: "Capitalize metadata keys") { context in
            context.markdownParser.addModifier(
                Modifier(target: .metadataKeys) { key, _ in
                    key.capitalized
                }
            )
        }
    }
}
```

Let me know what you think of that approach. I think having this customization be implemented as external plugins, rather than within Publish itself, makes more sense - and will be more flexible and easier to maintain in the long term.

JohnSundell:
The Ink changes were just merged into `master` by the way, if you want to give them a try to see if it'll work for your use case 👍

JohnSundell:
The Ink change I mentioned has now been released as part of Ink `0.3.0`, which the `master` branch of Publish now uses 👍

bellebethcooper:
Thanks @JohnSundell, I'll take a look at this again on the weekend and let you know how I go.

JohnSundell:
Sounds great 👍

bellebethcooper:
Had a look at this today, and it works perfectly with your plugin example above. Thanks very much @JohnSundell!

PR #45, opened 2020-01-12
[Draft] Introduce `.intermediate` folder and add support for `${variable}` substitution

ncreated:
Hello everyone 👋,

for my website needs I made a draft of variables substitution feature. I'm proposing this as `[Draft]` pull request, as I made significant changes to publishing pipeline and I'm not happy with the current shape of my code. It does the work for me, but it needs some polishing to be production ready 🔨. 

**I want to hear community feedback on this proposal.** I'd love to work more on this if you like it.  

I start with some examples of the Variable Substitution feature. Then I discuss modifications to `PublishingPipeline` in more details.

## Basic example

By specifying variable name, value and file extension:
```swift
.publish(
   // ...
   variables: .init(
      variables: ["author" : "Maciek"],
      in: ["md"]
   ),
   // ...
)
```
all occurrences of `${author}` will be replaced with my name in all `*.md` files during website generation.

Variables can be used in all files involved into website generation:
* content files,
* website resources,
* theme resources.

## More advanced example

For my website I use variables to configure website colors in Swift code. This includes styles defined in `styles.css` as well as icon templates rendered as `.svg` files:
```css
// styles.css

body {
    background: ${website-background-color};
    color: ${website-text-color};
}
```
```xml
// github-icon.svg

<svg role="img" ...>...<path fill="${website-text-color}" d="..."/></svg>
```
This is my `.publish()` method:
```swift
struct Colors {
   let background = UIColor(hex: "#ecf0f1")
   let text = UIColor(hex: "#ecf0f1")
}

variables: .init(
    variables: [
        "website-background-color": colors.background.hexString,
        "website-text-color": colors.text.hexString,
    ],
    in: ["css", "svg"]
),
```

💡 By using variables substitution, we can have more centralised control over the website (and it's content) purely in Swift code.

## Publishing pipeline modification

Current implementation of `PublishingPipeline` doesn't leave a room for files pre- or post- processing. All generation steps (`.copyResources()`, `.addMarkdownFiles()` and `.generateHTML()`) read content directly from website root folders (`/Resources`, `/Content`, theme resources) and save results to `/Output` folder.

Current pipeline:
```
/source-folders
↓
generationSteps() → /Output
```

To make the room for pre-processing, in this PR introduces `.intermediate` folder into publishing pipeline. All source files are initially copied to `.intermediate` folder. All generation steps read and write to `.intermediate` folder. At the end of the pipeline, content of `.intermediate/Output` is copied to final `/Output` folder:

```
/source-folders
↓
/.intermediate
↓
preProcessingSteps() -> /.intermediate
↓
generationSteps() → /.intermediate/Output
↓
postProcessingSteps() -> /.intermediate/Output
↓
/Output
```

This is how it looks in code:
```swift
.copyContentAndResourceFilesToIntermediateFolder(),
.copyThemeResourcesToIntermediateFolder(resources: theme.resources),

// ... pre-processing steps modifying`.intermediate` files

// ... generation steps reading from`.intermediate` and writing to `.intermediate/Output`

// ... post-processing steps modifying`.intermediate/Output` files 

.copyIntermediateOutputToFinalDestination(),
```

The `.substituteVariables(using:)` is one example of pre-processing step.

While I strongly believe this pipeline design unlocks many interesting use cases 🚀, I know it introduces semantic dependency between steps. It's incorrect to run steps from one group before running the other. This complexity might be mitigated by adding _pipeline validator_ or _steps dependency resolver_ components in the future.

## Implementation details

This PR brings three significant changes:
* It introduces `.intermediate` folder where all files are copied right after starting publishing pipeline.
* It adds new `.substituteVariables(using configuration: VariablesConfiguration)` step which can be applied to files in `.intermediate` folder.
* It updates generation steps (`.copyResources()`, `.addMarkdownFiles()` and `.generateHTML()`) to read files from `.intermediate` folder.

### DIFF
--- a/Sources/Publish/API/PublishingContext.swift
+++ b/Sources/Publish/API/PublishingContext.swift
@@ -34,7 +34,7 @@ public struct PublishingContext<Site: Website> {
     /// Any date when the website was last generated.
     public private(set) var lastGenerationDate: Date?
 
-    private let folders: Folder.Group
+    internal let folders: Folder.Group
     private var tagCache = TagCache()
     private var stepName: String
 
@@ -53,19 +53,24 @@ public struct PublishingContext<Site: Website> {
 }
 
 public extension PublishingContext {
+    func intermediateFolder(at path: Path) throws -> Folder {
+        do { return try folders.intermediate.subfolder(at: path.string) }
+        catch { throw FileIOError(path: path, reason: .folderNotFound) }
+    }
+
     /// Retrieve a folder at a given path, starting from the website's root folder.
     /// - parameter path: The path to retrieve a folder for.
     /// - throws: An error in case the folder couldn't be found.
-    func folder(at path: Path) throws -> Folder {
-        do { return try folders.root.subfolder(at: path.string) }
+    func sourceFolder(at path: Path) throws -> Folder {
+        do { return try folders.source.subfolder(at: path.string) }
         catch { throw FileIOError(path: path, reason: .folderNotFound) }
     }
 
     /// Retrieve a file at a given path, starting from the website's root folder.
     /// - parameter path: The path to retrieve a file for.
     /// - throws: An error in case the file couldn't be found.
-    func file(at path: Path) throws -> File {
-        do { return try folders.root.file(at: path.string) }
+    func sourceFile(at path: Path) throws -> File {
+        do { return try folders.source.file(at: path.string) }
         catch { throw FileIOError(path: path, reason: .fileNotFound) }
     }
 
@@ -73,7 +78,7 @@ public extension PublishingContext {
     /// - parameter path: The path to retrieve a folder for.
     /// - throws: An error in case the folder couldn't be found.
     func outputFolder(at path: Path) throws -> Folder {
-        do { return try folders.output.subfolder(at: path.string) }
+        do { return try folders.intermediateOutput.subfolder(at: path.string) }
         catch { throw FileIOError(path: path, reason: .folderNotFound) }
     }
 
@@ -81,7 +86,7 @@ public extension PublishingContext {
     /// - parameter path: The path to retrieve a file for.
     /// - throws: An error in case the file couldn't be found.
     func outputFile(at path: Path) throws -> File {
-        do { return try folders.output.file(at: path.string) }
+        do { return try folders.intermediateOutput.file(at: path.string) }
         catch { throw FileIOError(path: path, reason: .fileNotFound) }
     }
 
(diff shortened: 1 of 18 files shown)


JohnSundell:
Thanks for these suggestions, @ncreated.

When it comes to variable substitution, Publish (and Ink + Plot) was very much designed around avoiding string replacement, both for performance and type safety reasons. However, if you wish to implement string replacement for your own project, then I see no reason why it couldn't be done as a plugin rather than being built into Publish. And if that's currently not possible for some reason, then I think we should work on that, rather than including string replacement as a feature. In general, I'd like to keep the core of Publish as simple as possible, which is why it has such an extensible architecture - to enable external tools and plugins to be built around it.

Regarding your second idea, could you please elaborate a bit what you think are the use cases for pre- and post-processing within a Publishing pipeline? The only use case I've personally had for it so far is for caching, and there's already an API for that (which persists that kind of data between generation processes) - see `RSSFeedGenerator` for an example of how that's used. Again, I'm very reluctant to add this sort of complexity to Publish itself, unless there's a really good reason for it.

ncreated:
Thanks for the feedback @JohnSundell 💪. It confirms my anxiety of adding complexity to the pipeline internals and I totally get your point 👍.

When it comes to variables or string replacement, yes it is totally doable without touching the internals of `Publish`. As the exercise, I did it for my website and I needed only one small custom `PublishStep`: https://github.com/ncreated/n-created.com/commit/e0292845249b654a95010a713d5c9f0dcc6ad522. The difference is that I modify files directly inside `Output/` directory, not in `.intermediate/Output` folder as part of **atomic** publish process. So, if anything bad happens within my custom step my website is broken.

This isolation of publish pipeline was my main concern and reason to introduce `.intermediate` folder to let all steps operate on it without spoiling the `Output/` folder contents.

In other words, when it comes to files manipulation, current pipeline can be extended either on input or output side. Custom step or plugin can only modify source files (very bad idea) or output files (also bad idea if we want to guarantee Output correctness when pipeline fails).

> could you please elaborate a bit what you think are the use cases for pre- and post-processing within a Publishing pipeline?

Website resources processing which I did is only one example. Some other might include:
* Content generation / import - let's say I have a database with my products and I want to generate `*.md` posts for each. Or that I want my blog to mirror my Medium posts and download them during deploy.
* Resources fetching - like downloading website images so I can host them on my server.
* Links decoration - if I want to add / update affiliate information passed to my external links.

To be fair - all those examples are far beyond my need of having basic personal website 🙂.

JohnSundell:
@ncreated Regarding that all steps modify the Output directory directly, that's by design. Having `/Output` be a single source of truth reduces Publish's internal complexity, and makes it easier to both debug and get an overview of the current state of a website. It's true that an incorrect pipeline may produce an incorrect result, but then you just revert to the previous version, run again and you're good. I really don't see how adding an `.intermediate` folder adds any practical value, sorry.

And regarding pre- and post-processing, having a publishing step modify an output file isn't a bad idea at all - it's very much an intended use case. Again, having `/Output` be the working directory for all steps is by design. So your idea of pre- and post-processing can basically be implemented by just adding steps either before or after the steps that you want to pre/post process.

Finally, I'm not super interested in discussing adding new features based on hypothetical advanced use cases. Let's cross those bridges if we ever get there, otherwise we'll end up on a slippery slope towards over-engineering and unnecessary complexity.

So to me it seems like everything you want to accomplish is already possible, just in different ways than you might have originally thought of or prefer. I think we can close this one, do you agree? 🙂

JohnSundell:
And I'm sorry if my feedback comes off as being overly defensive about changing things in Publish. I'm not saying your ideas are bad. It's just that I've been maintaining OSS for a long time now, and any code that goes into the main project is code that I'll have to personally maintain and support, which is why I designed Publish to be super extensible - so that complexity can be distributed across the community through plugins and extensions. Which also gives you the freedom to implement things how you'd like 🙂

ncreated:
> Regarding that all steps modify the Output directory directly, that's by design. Having /Output be a single source of truth reduces Publish's internal complexity, and makes it easier to both debug and get an overview of the current state of a website. It's true that an incorrect pipeline may produce an incorrect result, but then you just revert to the previous version, run again and you're good.

I get this 👍 and cannot disagree in any way - when working on my changes, I saw the number of test branches to cover almost exploding 😉. 

I totally see your point of keeping the OSS projects simple and avoiding unnecessary complexity. Anyway, playing with the internals of `Publish` was super fun and I have my feature working on my website, so I'm totally okay with closing this PR. Thanks for you feedback @JohnSundell and see you around - I hope to help with sth on `Publish` in the future 🙂.

JohnSundell:
Awesome, thanks again for bringing this discussion up, and for contributing - I really appreciate it 😀 Hope to see you contribute again in the future 👍

PR #48, opened 2020-01-13
Clarify documentation around deploying

crazyroo1:
This clarifies the documentation around deploying. It states that deploying doesn't regenerate the site. 

Also, this adds a how-to for deploying on generation, with plenty of warnings.

### DIFF
--- a/Documentation/HowTo/deploying-on-every-generation.md
+++ b/Documentation/HowTo/deploying-on-every-generation.md
@@ -44,3 +44,5 @@ try MyWebsite().publish(using: [
 ```
 
 The `deployWhileGenerating(using: )` step will take in any defined deployment method, including the built in `.git()` method and any others you've defined customly. 
+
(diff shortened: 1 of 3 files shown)


JohnSundell:
Thanks for this @crazyroo1. However, I don't think that instructions on how to circumvent Publish's safety measures belongs in the official documentation. It's really not recommended to have a workflow like this, and even if we put a lot of caveats within the text, people will treat it as official instructions to "hack" `DeploymentMethod` to work like you described. So I'd be happy to merge a revised version of this documentation that's about "How to deploy your website" the official way, but it should be based on the best practices and conventions built into the framework, rather than circumventing them.

JohnSundell:
Any updates on this @crazyroo1? Otherwise I'll probably close this as I don't think it promotes best practices, as explained in my earlier comment.

crazyroo1:
Sorry about that delay, I didn't get any notification of a reply.

I definitely see where you're coming from, and agree. I'll close this and open a new pull request that just includes the clarification around using `--deploy`

PR #53, opened 2020-01-16
New article adding-color-to-tags

Ze0nC:
I made a new plugin for adding colors to tags by assigning the tags with different classes. 
(Here is an example on my site: https://czj.pw)

So I added an new how-to article about it. 

Cheers


JohnSundell:
Thanks for this @Ze0nC, but I don't think this warrants a "How to" article within the main Publish project. The "How to" articles are meant to explain how to use Publish's own features, not external plugins. I was happy to accept your previous PR about the Pygments plugin, since I think having different syntax highlighters easily available is a good thing, but I think this is too specific to be included in the main documentation. As long as you've added the `publish-plugin` tag to your project, then people should be able to eventually find it. So thanks anyway, but I'll close this.

Ze0nC:
Thanks for the reply. 
I now have a clearer understanding about the How-tos.

PR #55, opened 2020-01-18
Ensure that all images to which pages have relative links exist

wacumov:
This PR adds a new publishing step that checks whether all images to which pages have relative links exist. 

I am definitely going to add the tests, but I'm not sure where is the appropriate location for them. I suppose that `ErrorTests` is a good place for cases when `.ensureAllImagesExist()` step should fail, but I have no idea in which file "happy path" test case should be placed.

Also, I am not sure, should the default publishing pipeline contain this step.

And maybe this step should check not only images but local resources of other types too.

What do you think, @JohnSundell ?

### DIFF
--- a/Sources/Publish/API/PublishingStep.swift
+++ b/Sources/Publish/API/PublishingStep.swift
@@ -134,6 +134,23 @@ public extension PublishingStep {
             try MarkdownFileHandler().addMarkdownFiles(in: folder, to: &context)
         }
     }
+    
+    /// Ensure that all images to which items and pages have relative links exist.
+    static func ensureAllImagesExist() -> Self {
+        step(named: "Ensure all images exist") { context in
+            let content = context.getContent()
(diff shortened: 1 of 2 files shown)


JohnSundell:
Thanks a lot for this @wacumov, but I think it should be implemented as an external plugin instead. There's really no reason for the core Publish code base to contain this code, especially as the way that different people might want to verify their images might differ from project to project. If you implement this in a separate repo and add the topic `publish-plugin` then people should still be able to find it. Then you can also test all of your own logic within that separate project.

wacumov:
Thanks for the reply. Ok, I will implement this functionality as an external plugin in [this repository](https://github.com/wacumov/VerifyResourcesExistPublishPlugin.git) so I'll close the PR.

JohnSundell:
Sounds great! 👍

PR #57, opened 2020-01-19
Improve asSiteName extension

natanrolnik:
Hey John. First of all, thanks for all the contributions you've been doing for the community. You're truly an inspiration for many of us :hats-off:

This PR does the following:

* while trailing digits aren't valid (according to the existing tests I found), digits in the middle of the string should be valid in my opinion. For example: `Give2Charity` shouldn't have the `2` removed.
* when removing invalid characters, as dashes or underscores, the following character is capitalized. For example, `blog.somename.com` becomes `BlogSomenameCom`. The same is valid for a space or any other non-alphanumeric character.
* converts `asSiteName` to a variable, as it makes more sense in this case than a function

I've added a test and modified one existing to match the new behavior.

### DIFF
--- a/Sources/PublishCLICore/ProjectGenerator.swift
+++ b/Sources/PublishCLICore/ProjectGenerator.swift
@@ -19,7 +19,7 @@ internal struct ProjectGenerator {
         self.folder = folder
         self.publishRepositoryURL = publishRepositoryURL
         self.publishVersion = publishVersion
-        self.siteName = folder.name.asSiteName()
+        self.siteName = folder.name.asSiteName
     }
 
     func generate() throws {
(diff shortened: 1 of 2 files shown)


Sources/PublishCLICore/ProjectGenerator.swift
```diff
@@ -154,12 +154,21 @@ private extension Folder {
 }
 
 private extension String {
-    func asSiteName() -> Self {
-        let letters = filter { $0.isLetter }
-        guard !letters.isEmpty else {
+    var asSiteName: Self {
```
JohnSundell: This should not be converted to a property, since it performs `O(>1)` computation. Refer to [this article](https://www.swiftbysundell.com/articles/computed-properties-in-swift) for more detailed reasoning.

natanrolnik:
I would love to hear more of your thoughts on this. I usually add (computed) properties when the returned value is only derived from the class/type and is always the same, while I use functions when they actually create, change or trigger something.

I can revert that, no problem, but I’m just asking so maybe I could learn some reason why this is preferred.

JohnSundell:
@natanrolnik Did you read the article I linked you to? It explains my reasoning on this topic in great detail. Basically I think that a property should be `O(1)`, or at least very close to it, so that it can be accessed any number of times without causing additional computation. For example, compare `String.capitalized` (which is `O(1)`) with `String.uppercased()` (which is `O(N)`) in the Swift standard library. Since the `asSiteName` utility isn't `O(1)`, I believe it should be a method.

natanrolnik:
Oh, I opened the PR review on the GitHub app and your comment was hidden behind a tap, my bad 😄.
That makes total sense. Already pushed the fix. Thanks for the feedback.

PR #58, opened 2020-01-20
Determine which Python server command to use based on Python version

hisaac:
I currently have the `python` shell command bound to Python 3 on my system. The `SimpleHTTPServer` command has been deprecated in Python 3, in favor of `http.server` instead. This change calls the appropriate command depending on which version of Python is being run on the system.

Let me know if there are any changes or additions you'd like to see made. Thanks!

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -63,12 +63,12 @@ private extension WebsiteRunner {
         }
     }
 
-    func resolveSystemPythonMajorVersionNumber() -> Int? {
+    func resolveSystemPythonMajorVersionNumber() -> Int {
         // Expected output: `Python X.X.X`
         let pythonVersionString = try? shellOut(to: "python --version")
         let fullVersionNumber = pythonVersionString?.split(separator: " ").last
         let majorVersionNumber = fullVersionNumber?.first
-        return majorVersionNumber?.wholeNumberValue
+        return majorVersionNumber?.wholeNumberValue ?? 2
     }
 
     func outputServerErrorMessage(_ message: String) {
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -28,9 +28,19 @@ internal struct WebsiteRunner {
         """)
 
         serverQueue.async {
+
+            // If the system's Python version is Python 3, use the new `http.server` command.
+            // Otherwise, default to the `SimpleHTTPServer` command that Python 2 uses.
+            var pythonHTTPServerCommand: String
+            if self.systemPythonMajorVersionNumber == 3 {
+                pythonHTTPServerCommand = "http.server"
+            } else {
+                pythonHTTPServerCommand = "SimpleHTTPServer"
+            }
+
             do {
                 _ = try shellOut(
-                    to: "python -m SimpleHTTPServer \(self.portNumber)",
+                    to: "python -m \(pythonHTTPServerCommand) \(self.portNumber)",
                     at: outputFolder.path,
                     process: serverProcess
                 )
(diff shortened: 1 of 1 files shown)


Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -28,9 +28,19 @@ internal struct WebsiteRunner {
         """)
 
         serverQueue.async {
+
+            // If the system's Python version is Python 3, use the new `http.server` command.
```
JohnSundell: I think the code is self-documenting, and this comment is not needed.

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -28,9 +28,19 @@ internal struct WebsiteRunner {
...
+            // Otherwise, default to the `SimpleHTTPServer` command that Python 2 uses.
+            var pythonHTTPServerCommand: String
```
JohnSundell: Rather than including a local variable, I'd make the `systemPythonMajorVersionNumber` property a method instead called `resolvePythonServerCommand()` which simply returns the string directly. There's no need for any code outside of that method to be aware of the Python version, as far as I can tell.

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -55,6 +65,13 @@ private extension WebsiteRunner {
         catch { throw CLIError.outputFolderNotFound }
     }
 
+    /// The major version number of the current system's Python install (whatever Python responds to the `python` command)
+    var systemPythonMajorVersionNumber: Int? {
```
JohnSundell: Like mentioned above, I'd make this a method instead, since it actually performs XPC. Giving it property semantics sends the wrong message, IMO, since you'd assume that you can access it any number of times without a performance penalty, which isn't the case.

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -55,6 +65,13 @@ private extension WebsiteRunner {
         catch { throw CLIError.outputFolderNotFound }
     }
 
+    /// The major version number of the current system's Python install (whatever Python responds to the `python` command)
```
JohnSundell: No documentation needed here either, it isn't a public-facing API, and it should be very self-explaining.

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -55,6 +65,13 @@ private extension WebsiteRunner {
...
+    var systemPythonMajorVersionNumber: Int? {
+        let pythonVersionString = try? shellOut(to: "python --version")
+        let majorVersionNumberCharacter = pythonVersionString?.dropFirst(7).first
```
JohnSundell: I don't think it's a good idea to assume that the Python version string will always have this exact format. If they add a single whitespace or other character, this code will break. How about splitting the string by `" "` instead, and using the last component of that split to compute the version number?

Sources/PublishCLICore/WebsiteRunner.swift
hisaac: I like this. I've never really had a litmus test for when to use a variable versus a function for a simple method without arguments, but thinking of a performance penalty as the criteria is good. I'll make the change.

Sources/PublishCLICore/WebsiteRunner.swift
hisaac: Cool yeah, I didn't really like how I was doing this either, but couldn't think of a better option. Your idea to split it makes good sense. It still could break, but is probably less likely to break.

hisaac:
Thanks for the review @JohnSundell, I like what you've suggested. I'll make the changes ASAP.

hisaac:
OK, I made the recommended changes. Let me know what you think.

hisaac:
I also tried to write a test for this in `CLITests.swift`, but ran into an issue I don't understand.

I started by writing this:

```swift
func testWebsiteRunnerPythonServer() throws {
    let folder = try Folder.createTemporary()
    try makeCLI(in: folder, command: "new").run(in: folder)
    try makeCLI(in: folder, command: "run").run(in: folder)
}
```

When the test runs the `publish run` command, an exception is thrown by `serverProcess.terminate()` on line 48 of `WebsiteRunner.swift`, saying that it cannot terminate the process because the process wasn't started. Any suggestions on how to work around that for the test? Or perhaps a better question: Do you think a test is worth pursuing? If not, I'll just remove it.

Also, for future reference, I noticed that all the tests in `CLITests.swift` are behind build flags that are looking for `INCLUDE_CLI`, but I can't fine any other reference to `INCLUDE_CLI` in the project. How do you get those to run with that build flag?

JohnSundell:
@hisaac Testing the `run` command would be really difficult, as it starts the server process which blocks execution (by design). It's something we could look into in the future 🙂

hisaac:
Great, those changes look good to me. Thanks for the review, and thanks for Publish! I'm very glad to have made a contribution.

PR #62, opened 2020-01-23
Clean Output folder including hidden files

Hejki:
Hi, I want copy hidden files from Resources folder (for example: .htaccess) to Output folder. Problem is that `PublishingPipeline.setUpFolders` not removes hidden files.

This PR fix the cleaning step. If you agree I also add hidden files copying to copy file publishing steps.

Copy hidden files can be done in many variants:
* automatic copy hidden files by modify `folder.files` to `folder.files.includingHidden` in `PublishingStep.copyFiles`
* add parameter `includingHidde: Bool` to functions `PublishingStep.copyResources` and `PublishingStep.copyFiles`

### DIFF
--- a/Tests/PublishTests/Tests/FileIOTests.swift
+++ b/Tests/PublishTests/Tests/FileIOTests.swift
@@ -122,6 +122,19 @@ final class FileIOTests: PublishTestCase {
 
         XCTAssertEqual(itemFile?.name, "index.html")
     }
+
+    func testCleanHiddenFilesInOutputFolder() throws {
+        let folder = try Folder.createTemporary()
+        try folder.createFile(at: "Output/.hidden")
+        try folder.createFile(at: ".hidden").write("Hello, world!")
+
+        try publishWebsite(in: folder, using: [
+            .copyFile(at: ".hidden")
+        ])
+
+        let file = try folder.file(at: "Output/.hidden")
+        XCTAssertEqual(try file.readAsString(), "Hello, world!")
+    }
 }
 
 extension FileIOTests {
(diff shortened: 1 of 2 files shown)


artrmz:
How about making it configurable by adding a property to the `PublishingPipeline` struct? I am pretty sure some people might rather not want to skip hidden files.

Tests/PublishTests/Tests/FileIOTests.swift
```diff
@@ -122,6 +122,19 @@ final class FileIOTests: PublishTestCase {
 
         XCTAssertEqual(itemFile?.name, "index.html")
     }
+
+    func testCleanHiddenFilesInOutputFolder() throws {
+        let folder = try Folder.createTemporary()
+        try folder.createFile(at: "Output/.hidden")
+        try folder.createFile(at: ".hidden").write("Hello, world!")
```
JohnSundell: I think this test would be more robust if you didn't do this, and instead just asserted that the output folder *doesn't* contain any `.hidden` file at all. Right now you're relying on the `.copyFile` command throwing an error in case a file is already present, which is something that could change in the future, and would result in this test causing false positives.

JohnSundell:
@artrmz I don't think that's necessary. No-one should be assuming that anything gets persisted within the `Output` folder, even hidden files. That's what caching folders and custom folders are for.

artrmz:
@JohnSundell Yes, you are right. I thought at first (I guess it's too early in the morning yet 😅) this PR also includes changes for copying hidden files from Resources folder to Output. My bad.

Tests/PublishTests/Tests/FileIOTests.swift
Hejki: You are right. I updated it. Thanks for review.

PR #63, opened 2020-01-25
Move up the quick start for easier discoverability

pavopax:
I missed the quick start when this first came out, and yet it was the main thing I was looking for! Only found it again today.

I am sure others are also eager to quickly build a new site with Swift, so I encourage moving up the quick start to the top, like many other similar projects do.

Also added a "Learn more below" phrase to bottom of quick start

Thanks for this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,30 @@ Welcome to **Publish**, a static site generator built specifically for Swift dev
 
 Publish is used to build all of [swiftbysundell.com](https://swiftbysundell.com).
 
+## Quick start
+
+To quickly get started with Publish, install the command line tool by first cloning this repository, and then run `make` within the cloned folder:
+
+```
+$ git clone https://github.com/JohnSundell/Publish.git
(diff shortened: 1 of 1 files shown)


JohnSundell:
Thanks for contributing @pavopax, but I deliberately put "Quick start" towards the bottom of the README for a reason. It may seem like a strange reason at first; but I actually *want* people to read through the README before getting started with Publish, in order to understand some of the basics of how it works before using it. I know that a lot of developers these days are so eager to get started that with a project that they completely skip the documentation, but I really don't want my projects to be used that way, as it tends to lead to a lot of extra work for me down the line when those developers start asking me questions that the documentation already contains answers to. So thanks anyway, but I won't merge this in.

pavopax:
No problem man! Keep up the good work 👍

JohnSundell:
Thanks, and let me know if there are any other changes you'd like to see made 🙂

PR #69, opened 2020-01-31
WIP: Custom themes

daniloc:
Hey gang,

I'm in the process of building a custom theme for Publish and running into some friction I thought would likely be common. To that end, I've stubbed out the beginnings of a theming how-to.

Here are the places where I could use some backup on this:

# Extending `Website` using protocols

Imagine I have a theme that takes some input customization. A copyright string in the footer, user-defined colors, what have you. My assessment of the "proper" way to accept this input would be to add further `var`s on a package's `Website` struct. Does this sound right?

If so, and if I'm following @JohnSundell's intent, it seems that we'd then need to specify a theme-specific protocol for such `Website` implementations to adopt, and then use that as a type for the inputs of the HTML factory methods.

Again, does this sound right for this project? Conceptually all of this would make sense to me but I confess I'm a little stumped on the syntax 😅

Any syntax examples here would be most appreciated. I can then use them to flesh out the documentation stub.

Or if I'm chasing my tail with this approach, that's good feedback too! Let me know.

# Injecting custom style variables

Meanwhile, let's say I want to let users provide a handful of custom colors for various elements. What's the smartest way to get those into the output? I can imagine:

- Plugging a style string directly into a head tag
- Generating a CSS file from string interpolation
- Appending additional CSS elements to the theme's stylesheet (by including a publishing step?)

Appreciate input on the right approach here as well. And @JohnSundell, I know this was on your list of to-do's for the project, so if you already have a lot of this fleshed out in private and would prefer I just hold my horses until you can get to it, just say the word 😉

Thanks so much!


alexito4:
I can't speak for John's intend here but on my project I just built a theme that works specifically with my Website conforming type. I would expect that when I want to generalise that theme I would move the specific requirements into a specific them protocol that the website type could conform too. Sounds reasonable to me. 

No matter what I love the idea of adding some docs around this 👌

Documentation/HowTo/create-custom-themes.md
```diff
@@ -0,0 +1,68 @@
...
+            dependencies: ["Publish", "MyTheme"]
+        )
+    ]
+````
+
+Once added, Xcode will require you to edit the theme package from the same editor window as your website package.
+
+### Including stylesheets and other resources
+
+Use the `resourcePaths` argument when initializing your theme to ensure any necessary resources in your theme's package are included and copied to the `Output` folder:
```
doronkatz, line 48: Could you explain what this means @daniloc ?

doronkatz:
Hey guys. So how do I set an images folder resource and use it in this framework? @JohnSundell ?🙏🏼

PR #71, opened 2020-02-02
[PROPOSAL] CLI Arguments Revamp

MrSkwiggs:
## Proposal
Overhaul the CLI's API so as to define CLI Operations in a structured way. This will allow for better argument handling as well as offer more flexible Operation configuration.

## Motivation
The CLI's API is currently very limited. These changes aim to allow for deeper configuration on existing endpoints (`new`, `run`, ...) and for easy implementation of new endpoints, might that become necessary. My personal takeaway is that I wasn't able to name my Publish package any other way than naming the folder it was in. I want to be able to run `publish new MyCoolWebsite` in any folder I want.

## Current Changes
The API `CLI.run(in:)`'s underlying logic has been refactored to use a temporary `Operation` enum which we can switch on in `CLI.run(_:)`

excerpt:
```
internal enum CLIOperation {
    case new(args: Set<ProjectGenerator.Argument>)
    case generate
    case run(port: Int, args: Set<WebsiteRunner.Argument>)
    case deploy
}
```

These changes are non-breaking.

## Follow-up
Once this proposal is approved, I'll introduce the following protocols and refactor the current CLI Operations so that they conform to them.

### `CLIOperation`
Defines a CLI Operation, such as creating a new project, running an existing package, etc.
- `Argument` associated type, conforming to `CLIArgument`
- `RequiredArgument` associated type

### `CLIArgument`
Defines a type of argument that a given Operation requires/supports
- its keyword
- whether or not it is required for the given Operation
- its format, and how many parameters it requires

Once these protocols are implemented, I will be able to start writing generic logic that should allow to very easily tweak, configure & add new operations.

Furthermore, new tests for each `CLIOperation` implementation will be written, as well as making sure any existing tests don't break.

Looking forward to hearing back on this proposal :)


MrSkwiggs:
@JohnSundell Fair enough, thanks for taking the time to review this though 💪

PR #72, opened 2020-02-02
Add Output to .gitignore

dirtyhenry:
Hi!

I've just starting using Publish for a new project.

One of the first things I did was adding `Output` to the `.gitignore` since I don't really want generated content to hang out in the repository.

What do you think?

(as a next step, I realized that root folder names were a little all over the place using `String`, would it be worth moving all of those in a dedicated `enum` maybe?)

### DIFF
--- a/Sources/PublishCLICore/ProjectGenerator.swift
+++ b/Sources/PublishCLICore/ProjectGenerator.swift
@@ -47,6 +47,7 @@ private extension ProjectGenerator {
         /build
         /.build
         /.swiftpm
+        /Output
         /*.xcodeproj
         .publish
         """)


JohnSundell:
Hi @dirtyhenry, thanks for contributing. However, I don't think this is a good idea. If you git ignore the `Output` folder, then you won't be able to easily get an overview of the changes made to your site using `git diff`. I use that all the time to make sure that I didn't cause a regression with a recently introduced change, and to be able to see how my site has changed over time. After all, the output is arguably the most important aspect of your site, since that's what gets deployed to production. You're of course more than welcome to perform this change within your own projects, but I don't think it should be the default, so I'm closing this.

dirtyhenry:
Fair enough and actually great insight from your experience.

Hence, I followed up with: https://github.com/JohnSundell/Publish/pull/73

PR #74, opened 2020-02-06
Added GitHubPages DeploymentMethod

baguio:
Added a DeploymentMethod that publishes to GitHub with an extra dotfile that enables using Publish for GitHub Pages.
This also enables setting specific branches for Git and GitHub deployments

### DIFF
--- a/Sources/Publish/API/DeploymentMethod.swift
+++ b/Sources/Publish/API/DeploymentMethod.swift
@@ -36,37 +36,31 @@ public struct DeploymentMethod<Site: Website> {
 public extension DeploymentMethod {
     /// Deploy a website to a given remote using Git.
     /// - parameter remote: The full address of the remote to deploy to.
-    static func git(_ remote: String, branch: String = "master") -> Self {
+    /// - parameter branch: The branch to deploy to.
+    /// - parameter targetFolderPath: Any specific subfolder path to deploy the output to.
+    ///   If `nil`, then the output will replace all content in the branch.
+    static func git(_ remote: String, branch: String = "master", targetFolderPath: Path? = nil) -> Self {
         DeploymentMethod(name: "Git (\(remote))") { context in
-            let folder = try context.createDeploymentFolder(withPrefix: "Git") { folder in
-                if !folder.containsSubfolder(named: ".git") {
-                    try shellOut(to: .gitInit(), at: folder.path)
+            let folder = try context.createDeploymentFolder(withPrefix: "Git", targetFolderPath: targetFolderPath) { folder in
+                try folder.empty(includingHidden: true)
 
+                try shellOut(to: .gitInit(), at: folder.path)
+
+                try shellOut(to: "git remote add origin \(remote)", at: folder.path)
+
+                try shellOut(to: "git fetch", at: folder.path)
+
+                if targetFolderPath != nil {
                     try shellOut(
-                        to: "git remote add origin \(remote)",
+                        to: "git checkout \(branch) || git checkout -b \(branch)",
+                        at: folder.path
+                    )
+                } else {
+                    try shellOut(
+                        to: "git symbolic-ref HEAD refs/remotes/origin/\(branch)",
                         at: folder.path
                     )
                 }
-
-                try shellOut(
-                    to: "git remote set-url origin \(remote)",
-                    at: folder.path
-                )
-
-                _ = try? shellOut(
-                    to: .gitPull(remote: "origin", branch: branch),
-                    at: folder.path
-                )
-                
-                let gitFolder = try folder.subfolder(named: ".git")
-                
-                let subfolders = folder.subfolders.includingHidden
-                try subfolders.forEach { folder in
-                    if folder != gitFolder {
-                        try folder.delete()
-                    }
-                }
-                try folder.files.includingHidden.delete()
             }
 
             let dateFormatter = DateFormatter()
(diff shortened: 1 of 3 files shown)


baguio:
Oh, I forgot the documentation

JUSTINMKAUFMAN:
Nice! I had hacked together something to facilitate this (https://twitter.com/justinmkaufman/status/1230220460958158848?s=21) but your implementation is much better.

Sources/Publish/API/DeploymentMethod.swift
```diff
@@ -89,7 +89,75 @@ public extension DeploymentMethod {
...
+            try jekyllDisablingFile.delete()
+            
+            let ghPagesModeName : String
+            switch branch {
+            case .master : ghPagesModeName = "master branch"
+            case .masterDocs : ghPagesModeName = "master branch /docs folder"
+            case .ghPages : ghPagesModeName = "gh-pages branch"
+            }
+            
+            let settingsURL = "\(gitHubRemote(repository: repository, useSSH: false))/settings"
```
jackpal: The resulting URL is incorrect. It is:

 https://github.com/a/b.github.io.git/settings

should be:

 https://github.com/a/b.github.io/settings

(In other words you need to trim the ".git" from the url before adding the /settings.

jackpal: By the way, I ran into problems trying to use this patch.  The second time I tried to deploy to the same github project, the deploy step would produce an error because it was trying to add the .nojeckle file a second time.

Also I'm not sure that this extension is even necessary. I was able to use the regular github publisher to publish to my github.io page without any problem.

baguio: > (In other words you need to trim the ".git" from the url before adding the /settings.

Ok, changed the link generation so it doesn't insert the ".git" part every time. 

>  The second time I tried to deploy to the same github project, the deploy step would produce an error because it was trying to add the .nojeckle file a second time.

Right now checking on this one

> Also I'm not sure that this extension is even necessary

Not all repos use `master` for Github Pages, specially the ones generating a documentation site or a blog related to an already existing repo

baguio:
Ok, some changes:
- Fixed typo (oops) on the deployment method name message
- Fixed a broken link from the warning message for GitHub Pages repo setup
- Fixed a exception where dot files weren't being removed from the deployment folder, so deployments after the first one would fail because those files couldn't be replaced

john-mueller:
Thanks for working on this! I'll be using GitHub pages myself, so this is a great feature to have.

I did have quite a bit of trouble when testing it out. In particular, the original `git` deployment method is sensitive to whether or not you already have content on a branch, whether the branch exists on the remote, whether the .publish folder exists or not, etc. It also replaces the entire contents of a branch, so the master/docs source method currently doesn't embed the output *inside* the existing content as expected.

I fixed up the original git deployment method with some additional parameters, and added some documentation and tests. I've tested it extensively with all three Pages source methods. Give it a test, and if it works for you, you can just merge my [gh-pages-fixup](https://github.com/john-mueller/Publish/tree/gh-pages-fixup) branch on top of your commits.

Thanks again for getting this rolling!

bitomule:
Any updates on this? I'm interesting in publishing to github pages :)

baguio:
Ok, so i have to admit that this PR maybe should be split into two parts:

- One that extends its Git deployment capabilities (`branch`, maybe `targetFolderPath` from @john-mueller branch)
- One that adds GitHubPages deployment on top of those Git capabilities

i'll change this PR to be the first one. The other one will be created after this one is accepted

john-mueller:
> i'll change this PR to be the first one. The other one will be created after this one is accepted.
> 
> This should give these PR a proper goal since it currently changes Git methods and adds GitHubPages at the same time.

I don’t necessarily think it has to be split, since they are complementary additions. Given how busy @JohnSundell has been, and with how crazy the world is right now, having one less PR to review is probably good. I’d leave it as one unless he specifically requests a split. 🙂

JohnSundell:
Is this ready for me to review?

baguio:
@JohnSundell it is now

Sorry about the unnecessary wait. i ended up merging all changes into one PR because of not having that much time to follow what i said in the previous comment. So, @john-mueller, your changes entered this PR.

JohnSundell:
Thanks for this @baguio. However, after reviewing the code and thinking a lot about it, I've decided to not merge this PR into the main Publish repo, and would instead encourage you to create a separate repo that vends this deployment method as a plugin. My reasoning for that is:

A) I don't use GitHub Pages myself, and if I merge this code in, I'm also accepting the maintenance burden for it. That means that I'll likely end up having to update it as GitHub Pages changes its API and conventions, which is a responsibility that I don't wish to take on.

B) If I merge this in, I'm very likely to receive similar PRs for other hosting services (and there's a lot of them), which would be difficult to decline, given that Publish would then ship with GitHub Pages support built-in. Again, this would likely lead to an increased maintenance burden on my part, and would also lead to me having to maintain code that I have very little knowledge of.

Sorry if this seems overly conservative, but I spent a lot of time and effort in order to design the architecture of Publish to be highly modular, to enable code like this to live outside of the main repo - enabling other people to maintain such code (people who have the knowledge and motivation to do so). I'll gladly link to your plugin from Publish's README if you end up creating it though.

Thanks again for your work on this, but closing this PR for now.

baguio:
@JohnSundell completely agree with you. I tried it at first, but couldn't because a deployment method like this one needed a more complete Git method (custom branches and enabling custom dotfiles make this feature possible, custom folder path removes a limitation).
What do you think of updating this PR to insert these Git changes? I would move Pages to its own repo

JohnSundell:
Happy to hear that @baguio, and yeah, feel free to submit a new PR with anything you need to make this happen in an external repo, and I'll try to review those changes as fast as I can 👍

PR #75, opened 2020-02-25
Add ability to remove items from a section

peteschaffner:
This change adds a `PublishingStep` that makes it possible to conditionally remove items.

## Background

Coming from other static site generators, I have grown accustomed to writing posts labeled as drafts, which will be excluded from a build until marked as ready. This PR is meant to provide the capability to create such a workflow.

I use this feature for my site by adding an optional `draft: Bool?` item metadata property, and then conditionally run a step that removes such items from the build:

```swift
.if(CommandLine.arguments.contains("--removeDrafts"), .removeAllItems(matching: \.metadata.draft == true))
```

I then finish things off by passing a `--removeDrafts` flag when deploying my site.

### DIFF
--- a/Sources/Publish/API/PublishingStep.swift
+++ b/Sources/Publish/API/PublishingStep.swift
@@ -134,6 +134,26 @@ public extension PublishingStep {
             try MarkdownFileHandler().addMarkdownFiles(in: folder, to: &context)
         }
     }
+    
+    /// Remove all items matching a predicate, optionally within a specific section.
+    /// - parameter section: Any specific section to remove all items within.
+    /// - parameter predicate: Any predicate to filter the items using.
+    static func removeAllItems(
+        in section: Site.SectionID? = nil,
+        matching predicate: Predicate<Item<Site>> = .any
+    ) -> Self {
+        let nameSuffix = section.map { " in '\($0)'" } ?? ""
+
+        return step(named: "Remove items" + nameSuffix) { context in
+            if let section = section {
+                context.sections[section].removeItems(matching: predicate)
+            } else {
+                for section in context.sections.ids {
+                    context.sections[section].removeItems(matching: predicate)
+                }
+            }
+        }
+    }
 
     /// Mutate all items matching a predicate, optionally within a specific section.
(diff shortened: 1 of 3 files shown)


john-mueller:
👍 I think this is a great addition. Maybe you could also add your method for removing draft files to the /documentation folder.

peteschaffner:
> Very nice, thanks a lot for this @peteschaffner! Just made a quick little tweak to use the `removeAll(where:)` API instead of `filter` 👍

@JohnSundell: Ah, good call!

> 👍 I think this is a great addition. Maybe you could also add your method for removing draft files to the /documentation folder.

@john-mueller: Sounds like a good idea, but I'll let @JohnSundell decide if he would like to have this in the docs. In the meantime, you can checkout how I'm using this step [here](https://github.com/peteschaffner/peteschaffner.com/blob/master/Sources/PeteSchaffner/main.swift#L25).

alexito4:
I've used this on my website :) another data point that confirms this works 👍

pjechris:
That would be useful indeed :)

I came into a bug however: indexes are not rebuilt crashing my app. I made a fix of my own (https://github.com/pjechris/Publish/blob/master/Sources/Publish/API/Section.swift).

JohnSundell:
Ah that's a very good point @pjechris! Could you perhaps include that patch (https://github.com/pjechris/Publish/blob/master/Sources/Publish/API/Section.swift#L112) in this change, @peteschaffner, and add an additional test covering that the indexes have been rebuilt after an item was removed?

peteschaffner:
Ah good catch @pjechris! I included the patch and added an assertion to the existing test a la `testSortingItems`. @JohnSundell let me know if you you would prefer a separate test.

peteschaffner:
@JohnSundell I'm not sure what that `PublishTests.CLITests testProjectGeneration` failure is about, as all tests are passing for me. I wonder if it is a Swift version issue (5.2.2 on my machine and 5.1.3 on your Bitrise stack)?

JohnSundell:
And yeah, bumping the Swift version on Bitrise fixed the build issue. Will require Swift 5.2 from hereon out.

PR #77, opened 2020-03-11
ISO8601DateFormatter

dpfannenstiel:
As a web site developer I want to support time zone within my posts without having to write a custom date formatter. The default date formatter has support for the format "yyyy-MM-dd HH:mm", which ignores time zone.  

In Publish 0.5.0 date formatting is handled via the `DateFormatter` class, which conforms to the `Formatter` protocol.  Along with the many fine date formats there exists ISO 8601, which the Foundation framework implements in the `ISO8601DateFormatter` class, which also conforms to the `Formatter` protocol.  Since `PublishingContext` requires a `DateFormatter` I cannot use `ISO8601DateFormatter`.

This pull request makes two changes.

First, it introduces `AnyDateFormatter`, a minimal protocol to define a common interface for date formatters to conform.  Most of the API is already implemented by `DateFormatter` and  `ISO8601DateFormatter`, so implementing the protocol as an extension is trivial.

Second, `PublishingContext` changes  `dateFormatter` to a `AnyDateFormatter` and then make any subsequent changes to support that type change.  As part of this change, the default `dateFormatter` value is set to a `ISO8601DateFormatter`.  This will result in a breaking change for existing users of Publish, but it is believed that using a well known date format represents a common initial case.

### DIFF
--- a/Sources/Publish/API/AnyDateFormatter.swift
+++ b/Sources/Publish/API/AnyDateFormatter.swift
@@ -1,0 +1,28 @@
+//
+//  AnyDateFormatter.swift
+//  
+//
+//  Created by Dustin Pfannenstiel on 3/10/20.
+//  Copyright (c) Dustin Pfannenstiel, 2020
+//  MIT license, see LICENSE file for details
+//
+
+import Foundation
+
+public protocol AnyDateFormatter: Formatter {
+    var timeZone: TimeZone! { get set }
+    var dateFormat: String! { get set }
+
+    func date(from string: String) -> Date?
+    func string(from date: Date) -> String
+}
+
+extension DateFormatter: AnyDateFormatter {}
+
+@available(OSX 10.12, *)
+extension ISO8601DateFormatter: AnyDateFormatter {
+    public var dateFormat: String! {
+        get { "ISO8601DateFormatter" }
+        set { print("Warning: May not set dateFormat for ISO8601DateFormatter") }
+    }
+}
--- a/Sources/Publish/API/PublishingContext.swift
+++ b/Sources/Publish/API/PublishingContext.swift
@@ -22,7 +22,7 @@ public struct PublishingContext<Site: Website> {
     public var markdownParser = MarkdownParser()
     /// The date formatter that this publishing session is using when parsing
     /// dates from Markdown files.
-    public var dateFormatter: DateFormatter
+    public var dateFormatter: AnyDateFormatter
     /// A representation of the website's main index page.
     public var index = Index()
     /// The sections that the website contains.
--- a/Sources/Publish/API/PublishingContext.swift
+++ b/Sources/Publish/API/PublishingContext.swift
@@ -45,9 +45,15 @@ public struct PublishingContext<Site: Website> {
         self.folders = folders
         self.stepName = firstStepName
 
-        let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "yyyy-MM-dd HH:mm"
-        dateFormatter.timeZone = .current
+
+        let dateFormatter: AnyDateFormatter
+        if #available(OSX 10.12, *) {
+            dateFormatter = ISO8601DateFormatter()
+        } else {
+            dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd HH:mm"
+            dateFormatter.timeZone = .current
+        }
         self.dateFormatter = dateFormatter
     }
 }
--- a/Sources/Publish/Internal/MarkdownContentFactory.swift
+++ b/Sources/Publish/Internal/MarkdownContentFactory.swift
@@ -11,7 +11,7 @@ import Codextended
 
 internal struct MarkdownContentFactory<Site: Website> {
     let parser: MarkdownParser
-    let dateFormatter: DateFormatter
+    let dateFormatter: AnyDateFormatter
 
     func makeContent(fromFile file: File) throws -> Content {
         let markdown = try parser.parse(file.readAsString())
--- a/Sources/Publish/Internal/MarkdownMetadataDecoder.swift
+++ b/Sources/Publish/Internal/MarkdownMetadataDecoder.swift
@@ -11,12 +11,12 @@ internal final class MarkdownMetadataDecoder: Decoder {
     let codingPath: [CodingKey]
 
     private let metadata: [String : String]
-    private let dateFormatter: DateFormatter
+    private let dateFormatter: AnyDateFormatter
     private lazy var keyedContainers = [ObjectIdentifier : Any]()
 
     init(metadata: [String : String],
          codingPath: [CodingKey] = [],
-         dateFormatter: DateFormatter) {
+         dateFormatter: AnyDateFormatter) {
         self.metadata = metadata
         self.codingPath = codingPath
         self.dateFormatter = dateFormatter
@@ -73,11 +73,11 @@ private extension MarkdownMetadataDecoder {
         let keys: KeyMap<Key>
         let codingPath: [CodingKey]
         let prefix: String
-        let dateFormatter: DateFormatter
+        let dateFormatter: AnyDateFormatter
 
         init(metadata: [String : String],
              codingPath: [CodingKey],
-             dateFormatter: DateFormatter) {
+             dateFormatter: AnyDateFormatter) {
             self.metadata = metadata
             self.keys = KeyMap(raw: metadata.keys, codingPath: codingPath)
             self.codingPath = codingPath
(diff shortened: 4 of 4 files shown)


john-mueller:
I think the addition of `AnyDateFormatter` and ISO 8601 compatibility make sense. I don't think making it the default is the best option, however, as I think most people would rather use the current default "2020-03-12 15:00", as opposed to "2020-03-12T15:00:00-06:00".

You could use ISO 8601 with a custom step:

```
.step(named: "Use ISO8601DateFormatter") { context in
    if #available(OSX 10.12, *) {
        context.dateFormatter = ISO8601DateFormatter()
    }
}
```

It's probably worth adding that to [Documentation/HowTo/using-a-custom-date-formatter.md](https://github.com/JohnSundell/Publish/blob/master/Documentation/HowTo/using-a-custom-date-formatter.md).

dpfannenstiel:
I've updated the PR with better documentation on `AnyDateFormatter`.

My thoughts regarding the default date formatter are based around standards.  ISO 8601 has really wide adoption, whereas the current default is pretty custom and doesn't handle timezone at all.  I don't know what the roadmap to 1.0 looks like, but thinking about onboarding a new user you want to give them the minimum amount of friction to deployment.  So I guess it becomes a question of how important is time zone localization to a mature product for deployment.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Returns a date representation of a given string interpreted using the receiver’s current settings.
+    /// - parameter string: The string to parse.
+    func date(from string: String) -> Date?
+    /// Returns a string representation of a given date formatted using the receiver’s current settings.
+    /// - parameter date: The date to format.
+    func string(from date: Date) -> String
+}
+
+/// Apply `AnyDateFormatter` to `DateFormatter`
```
JohnSundell, line 28: This comment is unnecessary, since the code is completely self-explanatory.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+    var timeZone: TimeZone! { get set }
+    /// The date format string used by the receiver.
+    var dateFormat: String! { get }
+
+    /// Returns a date representation of a given string interpreted using the receiver’s current settings.
+    /// - parameter string: The string to parse.
+    func date(from string: String) -> Date?
+    /// Returns a string representation of a given date formatted using the receiver’s current settings.
+    /// - parameter date: The date to format.
+    func string(from date: Date) -> String
```
JohnSundell, line 25: This API is not used within Publish, so including it feels a bit like premature generalization, and makes the `AnyDateFormatter` protocol more complex than it has to be.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+
+import Foundation
+
+/// A common interface to abstract common methods and properties present
+/// in all date formatter types.
+public protocol AnyDateFormatter: Formatter {
+    /// The time zone for the receiver.
+    var timeZone: TimeZone! { get set }
+    /// The date format string used by the receiver.
+    var dateFormat: String! { get }
```
JohnSundell, line 18: This API is not used within Publish, so including it feels a bit like premature generalization, and makes the `AnyDateFormatter` protocol more complex than it has to be.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Returns a date representation of a given string interpreted using the receiver’s current settings.
```
JohnSundell, line 20: This documentation seems to be copied from Apple, which is not allowed. I'd recommend coming up with an original, simpler description of the method.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  AnyDateFormatter.swift
+//  
+//
+//  Created by Dustin Pfannenstiel on 3/10/20.
+//  Copyright (c) Dustin Pfannenstiel, 2020
+//  MIT license, see LICENSE file for details
```
JohnSundell, line 7: Please use the same start-of-file header as the rest of the project uses (but with your copyright, of course).

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+    func date(from string: String) -> Date?
+    /// Returns a string representation of a given date formatted using the receiver’s current settings.
+    /// - parameter date: The date to format.
+    func string(from date: Date) -> String
+}
+
+/// Apply `AnyDateFormatter` to `DateFormatter`
+extension DateFormatter: AnyDateFormatter {}
+
+/// Apply `AnyDateFormatter` to `ISO8601DateFormatter`
```
JohnSundell, line 31: Same thing here, this comment is not required and should be removed.

Sources/Publish/API/AnyDateFormatter.swift
```diff
@@ -0,0 +1,38 @@
...
+@available(OSX 10.12, *)
+extension ISO8601DateFormatter: AnyDateFormatter {
+    /// Since `ISO8601DateFormatter` does not have a variable format, return a descriptive string for reporting.
+    public var dateFormat: String! {
```
JohnSundell, line 35: I think a better solution would be to remove this from the protocol requirements, as recommended above.

Sources/Publish/API/PublishingContext.swift
```diff
@@ -45,6 +45,7 @@ public struct PublishingContext<Site: Website> {
         self.folders = folders
         self.stepName = firstStepName
 
+
```
JohnSundell, line 48: Please remove this whitespace.

JohnSundell (changes requested):
I think this is a great change, thanks @dpfannenstiel! I added a few comments inline that will need to be addressed before this can be merged in, but otherwise I'd be happy to make this a part of Publish's official API 👍

Sources/Publish/API/AnyDateFormatter.swift
dpfannenstiel: Digging into the PR notes I found a (better) version of this protocol exists as part of your [Codextended library](https://github.com/JohnSundell/Codextended/blob/8d7c46dfc9c55240870cf5561d6cefa41e3d7105/Sources/Codextended/Codextended.swift#L149). The major difference is that this version conforms to `Formatter`.  I'm pretty comfortable dropping that requirement.

I still think that `AnyDateFormatter` has a valuable place in *Publish* to support ISO 8601.  Including this protocol is redundant.  Not including it requires an `@import Codextended` anyplace we use it, regardless of other comments. I'd like your thoughts.

Sources/Publish/API/AnyDateFormatter.swift
dpfannenstiel: I had two reasons for including this.  Initially because in at least [one place](https://github.com/dpfannenstiel/Publish/blob/7cd7de754618a36af0c52fa33ae82077fa7e46d7/Sources/Publish/Internal/MarkdownMetadataDecoder.swift#L627) Publish reads a `dateFormat` out of a `AnyDateFormatter`.
The second reason is that from an API design perspective what `AnyDateFormatter` is doing is defining the minimum standard interface for a date formatter. A data formatter without read access to a date format didn't make sense to me.  If anything `ISO8601DateFormatter` represented a special case that needed to be handled. This way instead of having to special case date formatter every time we need the `dateFormat` we now know the property is available.

addisonwebb:
This is something I would love to have merged. Is there any way I can help out with this @dpfannenstiel?

Sources/Publish/API/AnyDateFormatter.swift
addisonwebb: This is actually used in `MarkdownMetadataDecoder` [here](https://github.com/JohnSundell/Publish/blob/6a2003f5ed2019cb2f566339f81f148726ce1b67/Sources/Publish/Internal/MarkdownMetadataDecoder.swift#L627).

dpfannenstiel:
@addisonwebb I’m trying to get @JohnSundell thoughts on how to include the AnyDateFormatter. Digging around I had found a version of it in a support library. But modifications are required. I’m not a huge fan of duplicating code needlessly and Codextended seems like the place to do it.

addisonwebb:
Ok @dpfannenstiel @JohnSundell I have submitted 2 PRs to address this issue using the existing `AnyDateFormatter` in Codextended.

- [Codextended PR #15](https://github.com/JohnSundell/Codextended/pull/15)
- [Publish PR #109](https://github.com/JohnSundell/Publish/pull/109)

These should be the minimal changes necessary to allow Publish users to use both `DateFormatter` and `ISO8601DateFormatter` as the custom date formatter on the publishing context. No breaking changes are introduced by these changes.

PR #78, opened 2020-03-12
Added title argument to PodcastFeedConfiguration

Goosse:
As it is, the title for a podcast is automatically set to the site's title.  This is undesirable for me as I develop a radio station website.  Multiple podcast feeds will be generated for different programs with different names.  So a title argument is necessary for each to have its own title.  

I would think that even in many cases where there is only one podcast feed for a website, the title for the podcast is likely not identical to the site title.


JohnSundell:
Hey @Goosse, thanks for this. In order to make this change backward-compatible, could you make `title` an optional parameter/property, and fall back to the site's title in case it's not defined? Otherwise this change will break all current usages, my own included.

Sources/Publish/Internal/PodcastFeedGenerator.swift
```diff
@@ -50,7 +50,7 @@ private extension PodcastFeedGenerator {
                   section: Section<Site>) throws -> PodcastFeed {
         try PodcastFeed(
             .unwrap(config.newFeedURL, Node.newFeedURL),
-            .title(config.title),
+            .title(config.title != nil ? config.title! : context.site.name),
```
grovdata, line 53: Perhaps use coalescing operator instead?
```
 .title(config.title ?? context.site.name),```

JohnSundell: Yeah, that'd be much better 👍

JohnSundell (commented):
Fantastic! Now all that we need is a test that verifies this fix and it can be merged in 👍

Goosse:
I finally just did this.  It got buried as a low priority issue, since I was using it on my fork for my website.  Hope this works.

Goosse:
I noticed you went through the pull requests recently.  Did you catch that I had added the test to this one?

PR #79, opened 2020-03-18
Add slash to links to avoid unnecessary redirects

marcprux:
Links to folder content generated by Publish don't seem to end with a trailing "/", which means there will be an additional HTTP redirect request to the resource with the path appended. You can see an example of this by going to https://swiftbysundell.com/articles/the-decade-of-swift in Safari: you will see it redirected to https://swiftbysundell.com/articles/the-decade-of-swift/, which is an unnecessary redirect.

This is more obvious when using curl:

```shell
marc@bix ~ % curl -v https://swiftbysundell.com/articles/the-decade-of-swift
  … blah blah blah …
< location: https://swiftbysundell.com/articles/the-decade-of-swift/
< 
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://swiftbysundell.com/articles/the-decade-of-swift/">here</a>.</p>
</body></html>
* Connection #0 to host swiftbysundell.com left intact
* Closing connection 0
```

TBH, I have no idea of this patch is the correct solution to this inefficiency, but in the absence of an issue tracker, there's no other forum for describing the issue.

### DIFF
--- a/Sources/Publish/API/PlotComponents.swift
+++ b/Sources/Publish/API/PlotComponents.swift
@@ -133,7 +133,7 @@ public extension Node where Context: HTMLLinkableContext {
     /// Assign a path to link the element to, using its `href` attribute.
     /// - parameter path: The absolute path to assign.
     static func href(_ path: Path) -> Node {
-        .href(path.absoluteString)
+        .href(path.absoluteString + "/")
     }
 }
 


grovdata:
Good catch, this seems like a bug to me.

bizz84:
Definitely! Not having trailing `/` can hurt SEO as well.

mwermeester:
Indeed, good catch!

mwermeester (approved):
Indeed, good catch!

mwermeester (approved):

steve-h:
I am converting my sites over to Publish. For now, I am using Publish as-is and overriding behaviour using added code and steps as intended by the Publish framework. I am going to use JSON-LD metadata on the pages and need URLs that are the true page reference.  Publish relies on redirects to slash terminated URL by the server. I used the  `python3 -m http.server` console output to look for all the ` HTTP/1.1" 301 -` redirects.

I then added; a clone of `Path .appendingComponent`
```swift
extension Path {
    func slashTerminated() -> Path {
        guard !string.isEmpty else {
            return ""
        }
        let separator = (string.last == "/" ? "" : "/")
        return "\(string)\(separator)"
    }
}
```

This is not necessarily efficient but should still work if Publish ever changes to add the slash.

I then do my own html generation. Fix the head canonical:
```swift
    return .head(
      .encoding(.utf8),
      .viewport(.accordingToDevice),
      .link(.rel(.canonical),
        .href(context.site.url(for: thisPath.slashTerminated()))),
```

Watch for the 301 redirects and change other `href` using routines like this on a cloned Theme:
```swift
 fileprivate static func itemList<T: Website>(for items: [Item<T>], on site: T) -> Node {
    return .ul(
      .class("item-list"),
      .forEach(items) { item in
        .li(
          .article(
            .h2(
              .a(
                .href(item.path.slashTerminated()),
                .text(item.title)
              )),
            .tagList(for: item, on: site),
            .p(.text(item.description))
          ))
      }
    )
  }
```

I have no problem with also doing the 301 redirects on the server for human users. My decision was swayed by JSON-LD and the Markdown relative links that truly need the trailing slash to be relative from.

tdc-peter (approved):

PR #80, opened 2020-03-29
RSS: Enable filtering of items using a predicate

JohnSundell:
This change enables users to decide whether to include a given item in a generated RSS feed using a predicate, which can be very useful when wanting to conditionally decide which items that go into a certain feed.


joeblau:
Do you have an example about how you would actually use this in a template's extension or in a website?

JohnSundell:
@joeblau Sure, here's how I used it when generating Swift by Sundell:
```swift
.generateRSSFeed(
    including: [.articles, .basics, .tips, .meta, .special, .sponsor, .podcast, .clips, .links],
    itemPredicate: Predicate {
        $0.sectionID != .special || $0.metadata.includeInRSSFeed == true
    }
)
```

PR #82, opened 2020-04-30
ctrl+c to stop

hishnash:
Ensure server process is also killed on `ctr+c`

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -24,8 +24,22 @@ internal struct WebsiteRunner {
         print("""
         🌍 Starting web server at http://localhost:\(portNumber)
 
-        Press ENTER to stop the server and exit
+        Press ENTER or CONTROL+C to stop the server and exit
         """)
+        
+        // Handle Ctrl+C shutdown
+        let signalsQueue = DispatchQueue(label: "Publish.signals")
+
+        let sigintSrc = DispatchSource.makeSignalSource(signal: SIGINT, queue: signalsQueue)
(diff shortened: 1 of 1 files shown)


Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -24,8 +24,22 @@ internal struct WebsiteRunner {
         print("""
         🌍 Starting web server at http://localhost:\(portNumber)
 
-        Press ENTER to stop the server and exit
+        Press ENTER or CONTROL+C to stop the server and exit
```
fishcharlie, line 27: So with this code. Does the server shut down gracefully when you hit enter? The comment below (on line 30) indicated that it does not. If that is the case and it does not exit gracefully on enter, I question if we should even keep the `Enter` feature. Seems like we should just replace `Enter` with `Control+C`. But might be a question for @JohnSundell. Just my two cents.

hishnash: Enter should work with the current version. If it does not then we need to look into some other solutio since all this pr does is ensure that ctrl+c does the same as enter

hishnash: for me this code work 90% of the time.. i have no idea why sometimes it does not.

JohnSundell: I don't think we should remove support for pressing `ENTER`, then we're just replacing one solution with another, which seems pointless. We should support both, and I'd love to find a way to ensure that this works 100% of the time before merging it in. After all, having an unreliable feature is worse than not having a feature at all.

PR #86, opened 2020-05-17
Extend Deployment Capabilities

baguio:
This is a continuation of PR #74 and following the plan outlined [here](https://github.com/JohnSundell/Publish/pull/74#issuecomment-627321541)

- Added a branch parameter to Git/GitHub deployment method
- Enabled usage of custom dot files inside deployment methods
- Added a custom path for deployment folder content ( thanks @john-mueller )

All of this should enable more possibilities for deployments and custom deployment methods

### DIFF
--- a/Sources/Publish/API/DeploymentMethod.swift
+++ b/Sources/Publish/API/DeploymentMethod.swift
@@ -57,8 +57,16 @@ public extension DeploymentMethod {
                     to: .gitPull(remote: "origin", branch: branch),
                     at: folder.path
                 )
-
-                try folder.empty()
+                
+                folder.includingHidden = true
+                
+                let gitFolder = try folder.subfolder(named: ".git")
+                var subfolders = try folder.subfolders.filter { (folder) -> Bool in
+                    folder != gitFolder
+                }
+                
+                try subfolders.delete()
+                try folder.files.delete()
             }
 
             let dateFormatter = DateFormatter()
--- a/Sources/Publish/API/DeploymentMethod.swift
+++ b/Sources/Publish/API/DeploymentMethod.swift
@@ -58,15 +58,15 @@ public extension DeploymentMethod {
                     at: folder.path
                 )
                 
-                folder.includingHidden = true
-                
                 let gitFolder = try folder.subfolder(named: ".git")
-                var subfolders = try folder.subfolders.filter { (folder) -> Bool in
-                    folder != gitFolder
-                }
                 
-                try subfolders.delete()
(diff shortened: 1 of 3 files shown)


Sources/Publish/API/DeploymentMethod.swift
```diff
@@ -36,29 +36,31 @@ public struct DeploymentMethod<Site: Website> {
...
+    /// - parameter branch: The branch to deploy to.
+    /// - parameter targetFolderPath: Any specific subfolder path to deploy the output to.
+    ///   If `nil`, then the output will replace all content in the branch.
+    static func git(_ remote: String, branch: String = "master", targetFolderPath: Path? = nil) -> Self {
         DeploymentMethod(name: "Git (\(remote))") { context in
-            let folder = try context.createDeploymentFolder(withPrefix: "Git") { folder in
-                if !folder.containsSubfolder(named: ".git") {
-                    try shellOut(to: .gitInit(), at: folder.path)
+            let folder = try context.createDeploymentFolder(withPrefix: "Git", targetFolderPath: targetFolderPath) { folder in
+                try folder.empty(includingHidden: true)
```
JohnSundell, line 45: This will break one of my plugins, which uses the git diff to determine what URLs to purge the Cloudflare cache for. In general, I don't think it's a good idea to completely reset the Git repository on every single deploy. Why is this needed?

baguio:
Sounds fair to me. Also, that would still need changes at PublishingContext

Sources/Publish/API/PublishingContext.swift
```diff
@@ -138,9 +138,12 @@ public extension PublishingContext {
     /// After that, all output files and folders will be copied into the new folder.
     /// - Parameter prefix: What prefix to apply to the folder, typically
     ///   the name of the current deployment method.
+    /// - parameter targetFolderPath: Any specific subfolder path to copy the output to.
+    ///   If `nil`, then the output will be copied to the deployment folder itself.
     /// - Parameter configure: A closure used to configure the folder.
     func createDeploymentFolder(
         withPrefix prefix: String,
+        targetFolderPath: Path? = nil,
```
JohnSundell, line 146: I think this should be named something like `outputFolderPath` instead, to clearly show that this path is used to determine where the output should be copied.

baguio:
@JohnSundell updated. Funny that it ended up being reduced to a couple lines

JohnSundell:
@baguio Made a few minor tweaks to the implementation, and added a test for it. Will merge as soon as CI passes 🙂

PR #88, opened 2020-05-28
CLI: Add plugin project generation command

JohnSundell:
This change makes it possible to run ‘publish new plugin’ in a folder to generate a plugin project. While plugins are just normal Swift packages, this makes it faster to get started building plugins.

The Swift tools version was also bumped to 5.2 for generated projects (to match Publish itself), and the minimum Publish version has been bumped to 0.6.0 (the latest release at this point).


oudomskn (commented):
Clear

PR #91, opened 2020-06-29
Add ability to specify branch for git and github

Kilo-Loco:
It would be nice to allow the user to specify a branch name during `git`/`github`.


omaralbeik:
@Kilo-Loco, @JohnSundell Now that Github uses `main` as the default branch name for new repos, do you think renaming the default value to `main` makes sense here?

PR #92, opened 2020-07-24
Add clean command to CLI

Kilo-Loco:
Occasionally `$ publish run` fails due to duplicate files inside the `.build` folder. This might be a problem on my machine since I have seen this behavior before. However, having a `clean` command would be nice in the event others are experiencing this and want to have a fresh build next time they run `$ publish run`.

### DIFF
--- a/Sources/PublishCLICore/CLI.swift
+++ b/Sources/PublishCLICore/CLI.swift
@@ -46,6 +46,9 @@ public struct CLI {
             let portNumber = extractPortNumber(from: arguments)
             let runner = WebsiteRunner(folder: folder, portNumber: portNumber)
             try runner.run()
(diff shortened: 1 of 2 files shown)


JohnSundell:
I'm very hesitant to accept this change since it makes assumptions about the way the Swift Package Manager is currently implemented (since it assumes that the private build folder will always be `.build`). I've made these kinds of assumptions in the past and they've been prone to break between different Swift versions, leading to an increased maintenance and compatibility burden. Also, since the Swift Package Manager has a built-in cleaning tool (`swift package clean`), I'm not sure why Publish would need to ship its own cleaning tool (given that it just uses SwiftPM to orchestrate its build process).

PR #94, opened 2020-08-24
Allow user to have draft content excluded from Output

Kilo-Loco:
It would be beneficial to be able to write up a draft and keep it in the project along with the published content. By setting `isDraft` to `true`, the content will not be included in the Output. The default value for `isDraft` is `false`.

```
article.md
---
date: 2020-08-24 09:00
description: My description
tags: mytag
image: images/image.jpg
isDraft: true
---
# Some title
```


antranapp:
Seems to be duplicated with https://github.com/JohnSundell/Publish/pull/89/files

JohnSundell:
Closed as a duplicate of https://github.com/JohnSundell/Publish/pull/89.

PR #95, opened 2020-09-06
Make `PublishingContext.copyFileToOutput` public

Szercinek12:
.


JohnSundell (changes requested):
Thanks for this @Szercinek12, but in order for this to be mergable you'll need to:

A) Implement tests for this change.
B) Follow the same indentation (4 spaces) as the rest of the project.
C) Add documentation, as all public APIs need to be properly documented.

JohnSundell:
Closing due to inactivity.

PR #96, opened 2020-09-25
Fix incorrect version in PublishCLI/main.swift

MaxDesiatov:
This file refers to Publish 0.6.0, which leads to generating new projects with templates that depend on 0.6.0. This doesn't seem to be right, as the latest version is 0.7.0.


JohnSundell (approved):
This doesn't really matter, as the Swift Package Manger will resolve the latest version up until `1.0.0`, but there's no harm in merging this in anyway, so let's go 🙂

PR #98, opened 2020-09-28
Fix horizontal scrolling in README.md

MaxDesiatov:
Current code formatting in the themes snippet requires horizontal scrolling. If it's slightly reformatted, horizontal scrolling can be avoided improving readability.


JohnSundell (approved):
Sure, why not? 🙂

PR #99, opened 2020-11-06
Adds possibility to change output path

trispo:
I needed a way to change the output folder to something like "Output/en" in order to do handle localized versions of my site.

### DIFF
--- a/Sources/Publish/Internal/PublishingPipeline.swift
+++ b/Sources/Publish/Internal/PublishingPipeline.swift
@@ -16,11 +16,12 @@ internal struct PublishingPipeline<Site: Website> {
 }
 
 extension PublishingPipeline {
-    func execute(for site: Site, at path: Path?) throws -> PublishedWebsite<Site> {
+    func execute(for site: Site, at path: Path?, outputPath: Path?) throws -> PublishedWebsite<Site> {
         let stepKind = resolveStepKind()
 
         let folders = try setUpFolders(
             withExplicitRootPath: path,
+            explicitOutputPath: outputPath,
             shouldEmptyOutputFolder: stepKind == .generation
         )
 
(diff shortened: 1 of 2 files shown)


Ze0nC:
Nice.
Actually I am working on adding the ability to generate multiple language website.
It is still in progress but you can check the current state: https://czj.pw
I am planning to release it later.

trispo:
Sounds great.

Sources/Publish/Internal/PublishingPipeline.swift
```diff
@@ -81,9 +82,10 @@ private extension PublishingPipeline {
     }
 
     func setUpFolders(withExplicitRootPath path: Path?,
+                      explicitOutputPath outputPath: Path?,
                       shouldEmptyOutputFolder: Bool) throws -> Folder.Group {
         let root = try resolveRootFolder(withExplicitPath: path)
-        let outputFolderName = "Output"
+        let outputFolderName = outputPath?.string ?? "Output"
```
JohnSundell, line 88: Since `outputPath` is a `Path`, this can no longer be treated as a folder name, and must also be treated as a `Path`. I suggest renaming this to `outputFolderPath`, and using `outputPath ?? "Output"` as its expression.

JohnSundell (commented):
I'm a little bit concerned about introducing this change to be honest, since I think there's a lot of power in the current convention that all Publish websites always output to the same `Output` folder. That also makes it possible for things like plugins and other tools used in conjunction with publish to assume that the output path will always be the same. Since you mentioned in the PR description that you're looking to add a nested `en` subfolder, could that be done by using a custom publishing step that moves all output files into that nested folder once you've finished generating your website instead?

Sources/Publish/Internal/PublishingPipeline.swift
trispo: Yes, that's what we did in the end. Let's close this PR. Thanks a lot.

PR #102, opened 2020-12-02
Change localhost server to SwiftNIO

mastermakrela:
# Replace python localhost server with Swift NIO one

## Advantages:
1. Publish CLI doesn't have to spawn python process
2. Publish CLI isn't dependent on python being available
3. Localhost server can be run through Xcode (earlier `readLine()` line didn't play nice with Xcode's console, see image below)

img1

4. We get solution to "ctrl+c to stop"(#82) for free

## Contribution
In this pull request I've updated the `/Sources/PublishCLICore/WebsiteRunner.swift` file to use Swift NIO server.
And I've written "How To" article about running through Xcode.

### _Disclaimers_
It's my first time contributing, so please correct me if I made something wrong 😅

_StaticServer Package is small, why didn't you integrate it directly into Publish?_
I know but for sake of modularity I thought it makes more sense to put this functionality into a separate Package.

---
Thank you for reading :)
Best Regards,
Christoph

### DIFF
--- a/Documentation/HowTo/running-localhost-server-in-xcode.md.md
+++ b/Documentation/HowTo/running-localhost-server-in-xcode.md.md
@@ -1,0 +1,30 @@
+#  Running localhost server in Xcode
+
+If you haven't already open package containing your website in Xcode.
+You can do it by either running `open Package.swift` in terminal or double clicking `Package.swift` in Finder.
+
+After Xcode fetches all package dependencies you should see the default build sheme.
+(It's the name of your website right to the run button on the toolbar.)
+
+## Creating run scheme
+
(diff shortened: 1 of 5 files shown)


JohnSundell:
Thanks for this @mastermakrela, and sorry for taking so long to provide any feedback on this. Unfortunately, I'm not interested in adding an external dependency to Publish, or to maintain a custom local server implementation. I think using Python for running a local server is perfectly fine, especially since the Publish CLI is just a convenience tool, it's not actually required to use Publish (so if users want to avoid using Python for some reason, they can just skip the CLI). Closing this PR.

PR #103, opened 2020-12-20
Added serve CLI utility

moonrise-tk:
Hello! As part of my website creation, I needed to do some post processing after Publish generates its html files. However, I also wanted to use Publish's built in server (`publish run`). I couldn't do this since `run` regenerates the website, deleting all of the post processing I had performed. Therefore, I added a "serve" CLI utility that just serves the code in the output directory without regenerating it. Therefore, my workflow would look like

```
publish generate
*postprocess*
publish serve
```

I moved all the `WebsiteRunner` class into `WebsiteServer` so now `WebsiteRunner` effectively just calls `WebsiteGenerator` then `WebsiteServer`.

### DIFF
--- a/Sources/PublishCLICore/CLI.swift
+++ b/Sources/PublishCLICore/CLI.swift
@@ -46,6 +46,10 @@ public struct CLI {
             let portNumber = extractPortNumber(from: arguments)
             let runner = WebsiteRunner(folder: folder, portNumber: portNumber)
             try runner.run()
+        case "serve":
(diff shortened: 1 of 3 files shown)


JohnSundell:
I think relying on external logic that isn't part of a website's publishing process is an anti-pattern, and shouldn't be encouraged by adding new CLI capabilities to support it. Any reason you can't perform your post-processing using a custom `PublishingStep`?

moonrise-tk:
Ah, you're right, I can.

PR #106, opened 2021-01-03
Generates AMP version of any location.

mvolpato:
I added to my own website the [AMP](https://amp.dev/documentation/guides-and-tutorials/start/create/?format=websites) version of all items.

I added a `AMPHTMLConfiguration` (similar to `TagHTMLConfiguration`), where you can declare which locations will have an AMP version. For instance

```swift
// Generate AMP pages, only for items.
var ampHTMLConfig: AMPHTMLConfiguration? = AMPHTMLConfiguration { location in
    if location is Item<YourWebsite> {
        return Path("\(location.path)/\(Path.defaultForAMPHTML).html")
    }
    return nil
}
```

Then in your implementation of `Theme`, you need to implement a new HTML generator for each type of location. If that specific location type (index, page, section, item, tag, and tag detail) is excluded by `AMPHTMLConfiguration`, then it does not matter what the related HTML generator will return.

The current version is not finished, but it is enough for my use case. If the community is interested in it (and if the way I implemented it is the preferred way), then I can finish it and complete the pull request. Given that this involves internal 
changes (mainly `HTMLGenerator.html`), I will wait for comments from the maintainers to go on with it.

Still TODO:

- [ ] Add generation of tag pages (not yet implemented); 
- [ ] Give basic implementation of the new HTML generators in `Theme+Foundation.swift`;
- [ ] Update the existing HTML generator in `Theme+Foundation.swift` and `PlotComponents.swift` to add the AMP rel link to the head section, only if the AMP version is generated.
- [ ] Tests
- [ ] Update README

### DIFF
--- a/Sources/Publish/API/AMPHTMLConfiguration.swift
+++ b/Sources/Publish/API/AMPHTMLConfiguration.swift
@@ -4,19 +4,24 @@
 *  MIT license, see LICENSE file for details
 */
 
-/// Configuration type used to customize how a website's
-/// AMP pages gets rendered. To use a default implementation,
+/// Configuration type used to customize which and where
+/// AMP pages are generated. To use a default implementation,
 /// use `AMPHTMLConfiguration.default`.
 public struct AMPHTMLConfiguration {
-    /// The path that will be appended to a resources to obtain the path of the AMP version.
-    public var suffixPath: Path
-
+    /// A closure that given a location, provides the path to the AMP version of that location.
+    ///
+    /// You can filter over the type of location (or a specific location) to return `nil`, which means that no AMP
+    /// version of that location is generated.
+    public var pathForLocation: (Location) -> Path?
+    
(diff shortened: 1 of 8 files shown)


JohnSundell:
Thanks for this contribution @mvolpato. To give you feedback, I think that this is a perfect example of the kind of functionality that should be implemented as an external plugin, rather than within Publish itself, given that AMP is a non-standard web technology. If you need any new (general-purpose) APIs in order to implement this as a plugin, then feel free to open separate PRs for those, but the changes outlined in this PR won't be accepted, as I'd like to keep the core of Publish as lightweight and simple as possible.

mvolpato:
Thanks for the feedback @JohnSundell

I will think about it. Probably not soon, unless there is demand from the community.

JohnSundell:
Cool, I'll close this PR in the meantime.

PR #107, opened 2021-03-12
[WIP] create release and build/upload binary for distribution

JulianKahnert:
In this PR I want to introduce a GitHub action that creates a release and builds/uploads the `publish` cli.

This [forums.swift.org](https://forums.swift.org/t/building-an-static-executable-for-distribution-on-macos/45497/6) thread helped me during development.

I would love to get your feedback here! 🙂

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -43,6 +43,7 @@ jobs:
           name: upload_url
       
       - shell: bash
+        id: get_upload_url
         run: |
           echo "::set-output name=upload_url::$(cat upload_url.txt)"
 
@@ -77,6 +78,7 @@ jobs:
           name: upload_url
 
       - shell: bash
+        id: get_upload_url
         run: |
           echo "::set-output name=upload_url::$(cat upload_url.txt)"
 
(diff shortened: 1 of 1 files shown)


JohnSundell:
Thanks @JulianKahnert, but I'm afraid that this not functionality that I'm interested in maintaining. It's easy enough for users to clone the project and run `make` to install it as a binary, so I don't really see much added benefit in having hosted binaries available. So, thanks anyway, but I'm going to close this.

JulianKahnert:
Hi @JohnSundell ,
totally understand your point of not wanting to maintain this kind of “feature”. Just to point out a use case:
https://github.com/JulianKahnert/blog/runs/3275430832?check_suite_focus=true

GitHub action doesn’t support macOS 11 yet, it is still in [preview](https://github.com/actions/virtual-environments#available-environments) for a limited number of testers.

And now my question:  do you have any experiences with shipping swift binaries that work with different swift runtimes? Is it really hard with swift and maybe not worth the time or is it a thing one (e.g. I) can/should have a deeper look at?

PR #114, opened 2021-07-17
Update Website.swift for sites without sections.

zenopolis:
Implementing a website without sections seems to be quite esoteric at the moment. This is an attempt to improve the situation, by providing a convenience implementation of WebsiteSectionID for just that purpose.


JohnSundell:
@zenopolis Thanks for this PR, and sorry for taking so long to provide feedback on it, but I don't think this is necessary to include as part of the public API. The recommendation for building single-section Publish websites would be to just define a single case (such as `main`) within your `SectionID` enum, since Publish was very much designed with sections in mind. If you want to do something drastically different from that section-based structure, then you'll probably be better off using Ink/Plot directly to generate your website. Closing this for now.

PR #115, opened 2021-11-05
Adds support for projects that contain multiple targets.

Zanchee:
You can now specify in the command line which target you wish to try and generate or run.


Zanchee:
Hey @JohnSundell! I've been using Publish over the past few days, but I hit a roadblock when I started supporting multiple targets inside my project. Hopefully this PR can solve that issue for others as well!

PR #117, opened 2021-11-08
Add initial support for Swift Concurrency

JohnSundell:
This change adds initial support for running certain publishing operations concurrently using Swift's new built-in concurrency system. This doesn't involve any breaking changes to the public API, apart from now requiring all Publish users to ensure that their HTML theme and multi-item mutation code can be executed in a concurrent manner.

**The following key changes were made as part of this:**

- Bump the project's Swift version to `5.5`, and update the CLI's project template accordingly.
- Add CollectionConcurrencyKit as a new package dependency, which adds concurrent versions of the various sequence mapping functions that Publish makes heavy use of.
- Make `PublishingStep` use an `async` closure type as its body, which in turn enables asynchronous calls to be made within any step, both built-in and custom ones.
- Each publishing pipeline is still performed in sequence, just like before. So it's only the internal operations within each step that may now be parallelized.
- The top-level `Website.publish` APIs are still completely synchronous, and uses an internal `DispatchSemaphore` to await the completion of the website's publishing pipeline before returning.
- A new built-in `mutateAllItems(in sections:)` API has been added, to make it convenient to iterate over all items within a certain set of sections in a concurrent manner.
- `HTMLGenerator` now executes concurrently, meaning that it can output a large number of HTML files in parallel, rather than writing all files one-by-one.
- `MarkdownFileHandler` now also runs concurrently, and processes all of its Markdown files in parallel, rather than doing so sequentially.
- `RSSFeedGenerator` and `PodcastFeedGenerator` now generates their respective feeds concurrently.

Note that this change is currently only compatible with macOS 12, as well as Linux, but that will be addressed before this new concurrency support is publicly released.

Overall, this set of changes results in a 4x speed improvement when generating Swift by Sundell using the M1 Max chip.


JohnSundell:
Feel free to try this out, and let me know how it works for your website. Here's how to do that:

1. Make sure that you're running macOS 12: Monterey, as the backward compatibility for Swift's concurrency features is still in beta as part of Xcode 13.2.

2. Replace the current Publish dependency declaration within your `Package.swift` manifest with the following:

```swift
.package(name: "Publish", url: "https://github.com/johnsundell/publish.git", branch: "swift-concurrency")
```

3. Generate your website just as you'd normally would. You shouldn't see any breaking API changes or regressions, but if you do, please report them by commenting on this PR. You should notice a significant speed increase, though, especially if your website contains a lot of items/pages.

bourvill:
Just tried,

do you have workaround for :
`error: the library 'SplashPublishPlugin' requires macos 10.10, but depends on the product 'Publish' which requires macOS 12.0`

JohnSundell:
@bourvill I just pushed my local SplashPublishPlugin changes that resolves that issue. Point your SplashPublishPlugin dependency to the following and you should be good to go:

```swift
.package(
    name: "SplashPublishPlugin",
    url: "https://github.com/johnsundell/splashPublishPlugin.git",
    branch: "swift-concurrency"
)
```

bourvill:
Works great !

insidegui:
Here are the results from testing on my website (rambo.codes) with the non-concurrent version vs. with the concurrent version. They show that even for a smaller website, there are significant improvements in speed (pretty much twice as fast with the concurrent version).

# MacBook Air 8-core M1 8 GB - macOS 12.0.1

## Baseline:

`0.26679396629333496`
`0.28175806999206543`
`0.26744401454925537`

## Concurrent version:

`0.1589679718017578`
`0.14613592624664307`
`0.16336405277252197`

# Mac mini 8-core M1 16 GB - macOS 11.6 (Xcode 13.2 beta 1) *

## Baseline:

`0.2709909677505493`
`0.27590394020080566`
`0.2848219871520996`

## Concurrent version:

`0.1625310182571411`
`0.16982603073120117`
`0.17325806617736816`

`*` In order to get things going on macOS versions prior to macOS Monterey, you'll have to manually copy `libswift_Concurrency.dylib` from within the Xcode 13.2 beta toolchain into the same directory where your executable is. You can find the dylib in `/path/to/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.5/macosx/libswift_Concurrency.dylib`.

If you don't do that, you'll get the following runtime error when attempting to run the executable:

```
dyld: Library not loaded: @rpath/libswift_Concurrency.dylib
```

I'm not sure if Apple is going to be able to easily address this for executables that are not distributed as a bundle (such as Mac/iOS apps). For apps, Xcode will automatically embed the concurrency library in the bundle.

PR #116, opened 2021-11-08
Checkout target deployment destination

schrismartin:
When using the `.git` or `.github` deployment strategies with a custom branch specified (for example, `gh-pages`), the deployment will fail with the following error message:
```
[info] error: src refspec gh-pages does not match any
error: failed to push some refs to 'github.com:my-name/my-name.github.io.git'
```

(This error can be seen in [this GitHub action](https://github.com/schrismartin/schrismartin.github.io/runs/4135033142?check_suite_focus=true#step:3:223))

This occurs because, when creating the `.publish/GitDeploy` directory, the HTML is generated in the default created branch (on some computers, `master`, and in others `main`) rather than the branch specified in the `branch` parameter. 

This PR adds a single additional command to checkout the specified branch pulled from the previous step. 

Alternatively, I considered calling `git init -b \(branch)` to initialize the internal repo with the intended branch. I chose the submitted approach instead because it doesn't require the implementer to remove their `.publish` directory for the change to apply.

PR #118, opened 2021-11-11
Fix tests when default git branch is not master

john-mueller:
## Issue

Two tests fail when git default branch name is not "master".

## Context

Since git version 2.28, the default branch name when creating a new repo can be set by adding 

```
git config --global init.defaultbranch branchname
```

When this has been set, `testGitDeploymentMethod()` and `testGitDeploymentMethodWithError()` fail, as they assume that the default branch is called "master".

## Resolution

Updated tests to use a deterministic branch name.

Also changed git `DeploymentMethod` to create the local repository using the branch name specified by the publishing step, rather than default branch name configured in git.

### DIFF
--- a/Sources/Publish/API/DeploymentMethod.swift
+++ b/Sources/Publish/API/DeploymentMethod.swift
@@ -41,7 +41,7 @@ public extension DeploymentMethod {
         DeploymentMethod(name: "Git (\(remote))") { context in
             let folder = try context.createDeploymentFolder(withPrefix: "Git") { folder in
                 if !folder.containsSubfolder(named: ".git") {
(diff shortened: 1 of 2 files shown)


john-mueller:
Apparently, bitrise has an older (<2.28) version of git on Linux, without the `-b`/`--initialbranch` flag on `git init`. I’ll fix it with a more compatible command tomorrow.

JohnSundell:
Thanks for this @john-mueller, and sorry for taking so long to provide any feedback on this. I just merged another PR that also addressed this very issue (https://github.com/JohnSundell/Publish/pull/116), so will close this one. Thanks again, and I really apologize for the lack of feedback.

PR #119, opened 2021-12-18
Podcast feed generator episode image feature

Goosse:
This pull request adds the functionality of using an item's imagePath to include an image unique to that episode in the RSS feed.


JohnSundell:
@Goosse Could you perhaps merge the latest `master` into this PR branch, to see if that fixes the CI issues?

Goosse:
Looks like the merge fixed the CI issues

Sources/Publish/Internal/PodcastFeedGenerator.swift
```diff
@@ -107,7 +110,7 @@ private extension PodcastFeedGenerator {
                     .summary(item.description),
                     .explicit(metadata?.isExplicit ?? false),
                     .duration(audioDuration),
-                    .image(config.imageURL),
+                    .image(imageURL ?? config.imageURL),
```
insidegui, line 113: Wouldn't it be better to exclude the `image` tag altogether when the item has no specific image?

As far as I've been able to tell, all clients will fall back to the main `<itunes:image>` at the channel level when the episode has no custom artwork. For shows with lots of episodes, having these duplicate tags in the feed add up and make the feed larger than it needs to be.

PR #120, opened 2021-12-31
Add HTML Mutator capability for Plugins.

dannys42:
Add HTML mutation capability to Plugins.  This allows Plugins to perform any number of last-minute changes on the raw HTML output.  Some examples:
 - HTML validation
 - pretty printing
 - minification
 - analytics tracking insertion


dannys42:
@JohnSundell Let me know what you think about this approach.  I'll add more comments to the public APIs if you think this is the right direction.  Let me know if you think it warrants any unit tests as well.

PR #121, opened 2022-02-08
Add Website option to ignore paths by name regex

atomicbird:
This adds an optional `ignoredPaths` array to `Website` which may contain a list of regular expressions of file or folder names that should be ignored during publishing. Each file or folder name is compared to entries in the array using `range(of: name, options: .regularExpression)`, so wildcards may be used. A structure that adopts `Website` may optionally declare something like this to affect the output:

```
    var ignoredPaths: [String]? { ["templates", "skip-this-file*"] }
```

My motivation for this is that I have a folder of template documents used on my site, which should not be copied to the site output.

### DIFF
--- a/Sources/Publish/API/PublishingContext.swift
+++ b/Sources/Publish/API/PublishingContext.swift
@@ -293,6 +293,7 @@ internal extension PublishingContext {
 
     func copyFileToOutput(_ file: File,
                           targetFolderPath: Path?) throws {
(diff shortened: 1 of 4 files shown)


atomicbird:
Pulling this back for now for some improvements, but I'll send it again soon.

atomicbird:
This PR now includes fixes for some problems I found that could cause unexpected regex matching, plus some tests.

atomicbird:
Replaced by #123

PR #122, opened 2022-02-21
Switch default branch from 'master' to 'main'

adamfootdev:
In line with Xcode and GitHub, I've updated the default branch to use 'main' instead of 'master' for git.


JohnSundell:
@adamfootdev As much as I'd love to change the default branch to `main`, this particular change that you're proposing is quite problematic, since it'll break all publishing workflows for users who haven't specified a branch manually (which I suspect is the vast majority of users). Instead, I think we should turn that branch argument into an optional, and detect whether to use `main` or `master` dynamically, to avoid breaking backward compatibility. Let me know if that's something you'd be interested in updating this PR to do, or if you'd like me to work on it.

PR #125, opened 2022-03-23
Switch python to python3.

moonagic:


keithcml:
I think this PR is essential for MacOS 12.3? `publish run` is failing as python not found.

moonagic:
> I think this PR is essential for MacOS 12.3? `publish run` is failing as python not found.

Yes, given that macOS has had Python3 builtin since several major version ago, I think this is by far the easiest way to do it.

JohnSundell:
Thanks a lot for this @moonagic! I made an extended version of this (https://github.com/JohnSundell/Publish/pull/132), which also cleans up the code that was previously used to dynamically resolve which Python version to use. Will merge that instead, so closing this. Thanks again!

PR #127, opened 2022-03-27
Ignore output folder

uhooi:


JohnSundell:
Sorry, but this change won't be accepted. You're free to git-ignore the `Output` folder locally if that's what you want, but I consider having the output checked into source control to be a vital regression-checking mechanism, since I'm able to see exactly what output changes that a source change resulted in.

PR #129, opened 2022-03-27
Keep resources folder

uhooi:
I want to prevent an error because `Resources` folder does not exist.

```
$ publish-cli generate
...
[step] Copy 'Resources' files
[path] Resources
[info] Folder not found
...
```

### DIFF
--- a/Sources/PublishCLICore/ProjectGenerator.swift
+++ b/Sources/PublishCLICore/ProjectGenerator.swift
@@ -62,7 +62,8 @@ private extension ProjectGenerator {
     }
 
     func generateResourcesFolder() throws {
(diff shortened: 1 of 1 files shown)


JohnSundell:
@uhooi By default, the `copyResources` step that's run when you use the `publish(withTheme:)` API is wrapped in an `.optional` step, making it not fail if it encountered an error. Since I'd rather not write unnecessary files onto the file system (and since most users are expected to actually put resources in their resources folder), then I'd suggest you either:

A) Put a `.gitkeep` file in your own Resources folder, to solve the problem you're facing. That doesn't require any changes to Publish itself.
B) Wrap your `copyResources` step in an `optional` step, just like what the default API does.

PR #132, opened 2022-04-20
CLI: Always use Python 3

JohnSundell:
Since Publish now requires macOS 12, we can safely require Python 3 when using the CLI's `run` command to start a local web server.

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -30,7 +30,7 @@ internal struct WebsiteRunner {
         serverQueue.async {
             do {
                 _ = try shellOut(
-                    to: "python -m \(self.resolvePythonHTTPServerCommand()) \(self.portNumber)",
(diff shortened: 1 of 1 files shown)


rogerluan:
@JohnSundell thanks for fixing this! 🙏 Would you mind publishing this over to homebrew so we can make use of it? Right now the latest version there is 0.8.0 so I'm unable to run it (even with an `alias python=python3`) 🙈 

```sh
🌍 Starting web server at http://localhost:8000

Press ENTER to stop the server and exit

❌ Failed to start local web server:
/bin/bash: python: command not found
```

JohnSundell:
@rogerluan I don't maintain the Homebrew support for Publish, I'm afraid (it was added by someone else). I recommend installing Publish manually, using `make`, since it's a completely self-contained package that can easily be built and installed using Swift Package Manager. That way, you can always use the latest version. Instructions on how to do that are in the README.

rogerluan:
Ooh perfect! Thank you so much @JohnSundell! 😁

PR #138, opened 2022-09-19
Nmccann/watch files

nmccann:
This introduces a new (very small) dependency to allow for optional file watching, which is just a lightweight wrapper around Apple's own file watching capabilities. This dependency imports Cocoa, though I've seen on another pull request that limiting to building on Apple machines isn't a deal breaker.

File watching is debounced to prevent excessive regeneration when making frequent changes, as well as to address the fact that file changes are a bit "noisy" - a single change can yield multiple events from the FileWatcher dependency. 

The current implementation is very opinionated - only the default Sources, Resources and Content folders are watched for changes, and the debounce duration cannot be changed. All of these could be made configurable, but I think that could be done in a follow up PR, should this one be accepted. (And for folder watching in particular, it would be helpful if the CLI used [swift-argument-parser](https://github.com/apple/swift-argument-parser) to make it easier to add support for new arguments)

There is one edge case that I haven't been able to figure out - changes to files in the `Resources` folder trigger more than one regeneration - it seems that the regeneration itself ends up counting as a change which puts it into a loop that it eventually gets out of after a few cycles. The same file in the Sources or Content folder doesn't cause this problem. I have a solution for this (using Apple's CryptoKit to generate SHA256 hashes of the changed files, and compare those before scheduling another regeneration), but I'd like to know whether or not introducing an additional dependency would be a problem.

### DIFF
--- a/Sources/PublishCLICore/WebsiteRunner.swift
+++ b/Sources/PublishCLICore/WebsiteRunner.swift
@@ -11,29 +11,41 @@ import FileWatcher
 
 internal struct WebsiteRunner {
     static let normalTerminationStatus = 15
-    static let debounceInterval: TimeInterval = 3
+    static let debounceDuration = 3 * NSEC_PER_SEC
     static let runLoopInterval: TimeInterval = 0.1
+    static let exitMessage = "Press CTRL+C to stop the server and exit"
     let folder: Folder
     let portNumber: Int
     let shouldWatch: Bool
 
+    var foldersToWatch: [Folder] {
+        get throws {
+            try ["Sources", "Resources", "Content"].map(folder.subfolder(named:))
+        }
+    }
+
     func run() throws {
-        var lastModified: Date?
-        var watcher: FileWatcher?
         let serverProcess: Process = try generateAndRun()
 
+        var watchTask: Task<Void, Error>?
+
         if shouldWatch {
-            watcher = try startWatcher {
-                if lastModified == nil {
-                    let file = try? File(path: $0)
-                    print("Change detected at \(file?.name ?? "Unknown"), scheduling regeneration")
+            watchTask = Task.detached {
+                for try await _ in changes(on: try foldersToWatch, debouncedBy: Self.debounceDuration) {
+                    print("Changes detected, regenerating...")
+                    let generator = WebsiteGenerator(folder: folder)
+                    do {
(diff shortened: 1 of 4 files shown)


Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -1,20 +1,117 @@
...
 import Foundation
 import Files
 import ShellOut
+import FileWatcher
 
 internal struct WebsiteRunner {
+    static let normalTerminationStatus = 15
+    static let debounceDuration = 3 * NSEC_PER_SEC
+    static let runLoopInterval: TimeInterval = 0.1
+    static let exitMessage = "Press CTRL+C to stop the server and exit"
```
nmccann, line 16: I was unable to get the folder watching to work while maintaining the `Press ENTER to stop the server and exit` behaviour - I replaced `readLine()` with a non-blocking implementation (checking standard input on each iteration of the `while` loop), but that interfered with the watching functionality for an unknown reason.

Sources/PublishCLICore/WebsiteRunner.swift
```diff
@@ -1,20 +1,88 @@
...
 import Foundation
 import Files
 import ShellOut
 
+#if canImport(FileWatcher)
+import FileWatcher
+#endif
+
 internal struct WebsiteRunner {
+    static let nanosecondsPerSecond: UInt64 = 1_000_000_000
```
nmccann, line 16: Workaround for `NSEC_PER_SEC` not being available on Linux which according to [this](https://forums.swift.org/t/nsec-per-sec-is-an-unresolved-identifier-on-linux/34991) is a bug. Alternatively, I could have wrapped the NSEC_PER_SEC usage in a `#if canImport(FileWatcher)` condition, since it's only used alongside FileWatcher.

nmccann:
The current Linux failure (in `RSSFeedGenerationTests.swift`) appears to be unrelated to my changes - I am unable to locally reproduce that failure (when, for instance, running on a linux Docker container). I see there is a similar (though in a different file) failure on the main branch [here](https://app.bitrise.io/build/c24c826b-4200-4149-ae91-7f5881fce198) - perhaps the affected tests are flakey?

mwermeester:
Agree, can somebody have a look at the Linux build failures?

PR #141, opened 2022-12-14
Make publish function async

alexito4:
The internal semaphore used to make the CLI wait for publish to finish is no longer necessary and causes problems with concurrency. This PR removes it and makes the publish function async.

- Is no longer necessary because Swift already supports async main, so one can build CLIs that use async and automatically await for the work to finish. 
- It cause problems because is blocking the main thread. If any async code needs to run on the MainActor this causes a deadlock.


JohnSundell:
Thanks a lot for this PR, @alexito4, and I'm so sorry for the lack of feedback until now. I just merged in an alternative implementation (which I saw that you also have a 👍 to) in PR #143. I think, for now, we want to maintain backward compatibility with non-`async` call sites, but eventually we should deprecate those semaphore-based APIs, and only keep the `async` versions. But for now, I'll close this PR. Thanks again!

PR #142, opened 2023-01-02
Validate HTML generator produces unique files

erichoracek:
If you have a website with a `SectionID` case that has the same name as a page markdown file, right now whichever one ends up in the output is nondeterministic. This change adds a validation step to the `HTMLGenerator` to ensure uniqueness.


codewithhart:
Interesting find. I just spent way too long figuring out why the top-level *.md files weren't updating. From what I've seen around the web...is the correct behavior to have a SectionID case match the .md file?

PR #143, opened 2023-03-12
Add public async versions of `publish` method

yonihemi:
## Motivation
- Allow Publish to run on [platforms](https://swiftwasm.org) which cannot support a `DispatchSemaphore`.
- Allow calling Publish from async contexts without the risk of deadlock.
- Keeping existing clients intact.

## Changes
Adding alternative `async` versions of the `Website.publish` methods.

Compared with the approach of [replacing sync with async versions](https://github.com/JohnSundell/Publish/pull/141), this PR keeps all tests passing and *almost* doesn't break calling sites.

This is technically a breaking change but in a pretty minimal way, according to calling context:
- In a `synchronous` function, the sync version will be inferred as before - no change.
- In `main.swift` top-level code, the sync version will be inferred as before (adding `await` in the call will result in a warning).
- In an `async` function - the code will break with a helpful compiler error:
```
Expression is 'async' but is not marked with 'await' 
Fix: Insert 'await '
```

### DIFF
--- a/Sources/Publish/API/Website.swift
+++ b/Sources/Publish/API/Website.swift
@@ -130,6 +130,67 @@ public extension Website {
         semaphore.wait()
         return try result!.get()
     }
+    
+    /// Publish this website using a default pipeline. To build a completely
+    /// custom pipeline, use the `publish(using:)` method.
+    /// - parameter theme: The HTML theme to generate the website using.
+    /// - parameter indentation: How to indent the generated files.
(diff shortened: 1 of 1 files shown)


JohnSundell (approved):
Thanks a lot for this PR, @yonihemi! I think this is a really elegant solution, which (as you noted) provides the best amount of backward compatibility that can be achieved in this case 👍

PR #151, opened 2023-11-24
Feature: Page metadata

jhonatn:
Reuses the implementation of WebsiteItemMetadata to add metadata to pages, but adds a default type for the PageMetadata associatedtype so it is not a hard requirement


jhonatn:
One test is failing, but it appears like the base commit in master was also failing, and reading the test doesn't give me much clue on how that method was supposed to throw an error

